PR #41, opened 2019-05-14
Update Structure section with new sub-repo names

rzadp:


rzadp:
Using admin to merge without approve because it's trivial

PR #51, opened 2019-05-20
Fix the docker-related logrotation config

ArseniiPetrovich:
From Ansible README:
For those used to `/usr/bin/chmod` remember that modes are actually octal numbers. 
You must either add a leading zero so that Ansible's YAML parser knows it is an octal number (like 0644 or 01777), or quote it (like '644' or '1777') so Ansible receives a string and can do its own conversion from string into number.

The error was in using `644` mode instead of `0644`

### DIFF
--- a/deployment/oracle/roles/logging/tasks/main.yml
+++ b/deployment/oracle/roles/logging/tasks/main.yml
@@ -23,7 +23,7 @@
     dest: /etc/rsyslog.d/30-docker.conf
     owner: root
     group: root
-    mode: 644
+    mode: 0644
 
 - name: Set the log configuration file to send container logs to remote server
   template:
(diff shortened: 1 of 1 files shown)


rzadp:
I've configured CircleCI to run on forked requests but apparently it will work only on newly created PRs so merging without CI checks.

MarkusTeufelberger:
https://docs.ansible.com/ansible-lint/rules/default_rules.html

E202

In case you want to prevent this in the future or check your other repositories...

rzadp:
> https://docs.ansible.com/ansible-lint/rules/default_rules.html
> 
> E202
> 
> In case you want to prevent this in the future or check your other repositories...

Should be added in #36

Issue #53, opened 2019-05-21
Update oracle-e2e to work with contracts at 2.3.0-rc0


rzadp:
Currently they only work with with `2.2.0`

akolotov:
Details what it is necessary to be done to migrate e2e testing to the next version of contracts provided here: https://github.com/poanetwork/tokenbridge/pull/54#issuecomment-494788260

akolotov:
The [new release v2.3.0](https://github.com/poanetwork/poa-bridge-contracts/releases/tag/2.3.0) of the TokenBridge contracts has been released

Issue #52, opened 2019-05-21
Update ui-e2e to work with contracts at 2.3.0-rc0


rzadp:
Currently they only work with with ~~`2.1.0`~~ `2.2.0` as of #54

akolotov:
Details what it is necessary to be done to migrate e2e testing to the next version of contracts provided here: https://github.com/poanetwork/tokenbridge/pull/54#issuecomment-494788260

akolotov:
The [new release v2.3.0](https://github.com/poanetwork/poa-bridge-contracts/releases/tag/2.3.0) of the TokenBridge contracts has been released

Issue #44, opened 2019-05-14
Remove dependency on `package-lock.json` in Dockerfiles


rzadp:
The Dockerfiles are COPYing `package-lock.json` when building images.
We should remove this dependency, or rely on `yarn.lock` instead.

rzadp:
### Dependencies problem

The Docker build processes rely on sub-repository `package.lock` and `package-lock.json`when installing dependencies.

Our local build process installs dependencies for all sub-repositories at one time, and uses one `yarn.lock` in root directory.

We shouldn't have different methods of installing dependencies locally and when building docker images because it can cause tricky errors.

### Number of Dockerfiles

There is a total of 7 Dockerfiles, plus one in `poa-bridge-contracts`.

![Screenshot 2019-05-14 at 13 06 20](https://user-images.githubusercontent.com/12039224/57694217-6d4e8d00-764b-11e9-9ab6-08301791a6de.png)

Seems excessive and increases complexity.

### Proposed solution

I'm thinking about having one Dockerfile in root directory. It would install dependencies for sub-repositories using main `yarn.lock`, and would be used as a build base for sub-repositories `docker-compose` files.

@akolotov @phahulin @ArseniiPetrovich Do you have any comments on this?

akolotov:
@patitonar could you take a look?

akolotov:
My understanding is that some of these docker files are used for e2e testing only, whereas others are to build images used in production. @patitonar please correct me if I am wrong.

patitonar:
> We shouldn't have different methods of installing dependencies locally and when building docker images because it can cause tricky errors.

I agree on this, we should use `yarn` only.

Regarding the Dockerfile most of them are use for e2e testing and can be replaced by accessing the sub-repository needed or be unified in the case of the repeated Parity nodes Dockerfile used on e2e tests. I think the only one used for build for production along with the docker-compose file are the one from the oracle sub-repository

Issue #21, opened 2019-05-07
Extract repeated submodules/poa-bridge-contracts into one in root dir


rzadp:
Release 2.3.0-rc0 should be used for poa-bridge-contracts.

akolotov:
I would add even more - instead of usage of ABI sourced in the `token-bridge/abis` directory it is necessary to compile them from the contract submodule. By doing this we will always have updated ABIs without necessity to update them manually every time when the contracts are being changed.

akolotov:
In order to not mix the work let's consider this issue to
- [ ] introduce the git submodule for the bridge contracts in the root dir
- [ ] change the oracle and UI projects as so this submodule is used for e2e tests (this is only the place where the contract submodule is used now).

rzadp:
Fully finishing this is blocked on #52 and #53

PR #54, opened 2019-05-21
Remove sub-dockerfiles and sub-contracts

### LINKED ISSUE #6: Extract bridge-ui/e2e and configure CircleCI
- [x] Extract `bridge-ui/e2e`
- [ ] Run them on CI

rzadp:
### Purpose
The purpose of this PR is to:
- Remove dependency on `package-lock.json` and get rid of excessive `Dockerfiles` as per #44 
- Migrate into common `contracts` in root directory instead of nested submodules and `git clones` in sub-Dockerfiles as per #21 

### Achieved work
- `oracle-e2e` and `ui-e2e` share parity config in root directory
- `oracle-e2e` uses contracts in root directory
- `ui-e2e` uses contracts in root directory
- Reduced number of `Dockerfiles`
- Made `ui-e2e` work on CI - closes #6 

### Contracts issues
- `contracts/package.json` is installed by yarn and included in yarn workspaces, but `contracts/deploy/package.json` is installed separately by npm
  - yarn does not allow nested `package.json's`
- There are `contracts-2.1.0` ~~and `contracts-2.2.0`~~ because `oracle-e2e` and `ui-e2e` were hardcoded to use these tags and do not work with our `contracts` at `2.3.0-rc0`
  - Created #52 and #53 to fix this, then we can remove `contracts-*` and #21 can be completely finished
  - yarn workspaces don't work with sub-folders of the same name `poa-parity-bridge-contracts`, so the excessive sub-folders are handled by npm


Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM node:10
```
akolotov: seems that this `Dockefile` is used for e2e testing only. Please consider to have a separate directory `e2e` which will be a root for everything that is related to e2e testing. So this file will located there as well as the directories `parity`, `ui` (were `ui-e2e`) and `oracle` (were `oracle-e2e`).

oracle-e2e/deploy.js
```diff
@@ -2,7 +2,7 @@ const path = require('path')
 const shell = require('shelljs')
 
 const envsDir = path.join(__dirname, 'envs')
-const deployContractsDir = path.join(__dirname, 'submodules/poa-bridge-contracts/deploy')
+const deployContractsDir = path.join(__dirname, '../contracts-2.2.0/deploy')
```
akolotov: does it make sense to introduce a module with the constant to keep the name of the directory containing the bridge contracts as so this module and this constant will be used in other files: `oracle-e2e/scripts/deployERC20.js`, `oracle-e2e/test/ercToErc.js`, `oracle-e2e/test/ercToNative.js` and `oracle-e2e/test/nativeToErc.js`. So, later when the directory with contracts is renamed it will be necessary to change one file instead of five files.

ui-e2e/scripts/deploy.js
```diff
@@ -0,0 +1,22 @@
+const path = require('path')
+const shell = require('shelljs')
+
+const envsDir = path.join(__dirname, '../envs')
+const deployContractsDir = path.join(__dirname, '../../contracts-2.1.0/deploy')
```
akolotov: does it make sense to introduce a module with the constant to keep the name of the directory containing the bridge contracts as so this module and this constant will be used in another file `ui-e2e/scripts/deployERC20.js`. So, later when the directory with contracts is renamed it will be necessary to change just one file.

akolotov:
@patitonar do you have understanding why the e2e on UI side does not work with the contracts `v2.2.0`? I see that the difference between `v2.1.0` and `v2.2.0` is almost nothing from the UI point of view.

rzadp:
> @patitonar do you have understanding why the e2e on UI side does not work with the contracts `v2.2.0`? I see that the difference between `v2.1.0` and `v2.2.0` is almost nothing from the UI point of view.

I didn't spend time to look into this yesterday as I wanted to roll out the PR for review sooner. Maybe I'll be able to figure this out today.

oracle/README.md
```diff
@@ -51,11 +51,11 @@ For more information on the Redis/RabbitMQ requirements, see [#90](/../../issues
 
 ## Installation and Deployment
 
-**Note:** The following steps detail the bridge deployment process for development and testing. For deployment in a production environment we recommend using the [Bridge Deployment Playbooks](https://github.com/poanetwork/deployment-bridge/tree/master/bridge-nodejs). 
+**Note:** The following steps detail the bridge deployment process for development and testing. For deployment in a production environment we recommend using the [Bridge Deployment Playbooks](../deployment/README.md). 
 
 #### Deploy the Bridge Contracts
 
-1. [Deploy the bridge contracts](https://github.com/poanetwork/poa-bridge-contracts/blob/master/deploy/README.md)
+1. [Deploy the bridge contracts](../contracts/README.md)
```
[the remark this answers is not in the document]
rzadp: Done.

akolotov:
> > @patitonar do you have understanding why the e2e on UI side does not work with the contracts `v2.2.0`? I see that the difference between `v2.1.0` and `v2.2.0` is almost nothing from the UI point of view.
> 
> I didn't spend time to look into this yesterday as I wanted to roll out the PR for review sooner. Maybe I'll be able to figure this out today.

I think I found the difference. The following must be updated in `ui-e2e/envs/contracts.env`, `ui-e2e/envs/erc-contracts.env`, `ui-e2e/envs/erc-native-contracts.env`:
* `DEPLOYMENT_GAS_LIMIT=5000000`
* `HOME_OWNER_MULTISIG`, `HOME_UPGRADEABLE_ADMIN_VALIDATORS`, `HOME_UPGRADEABLE_ADMIN_BRIDGE` must be updated to `HOME_BRIDGE_OWNER`, `HOME_VALIDATORS_OWNER` and `HOME_UPGRADEABLE_ADMIN`
* `FOREIGN_OWNER_MULTISIG`, `FOREIGN_UPGRADEABLE_ADMIN_VALIDATORS`, `FOREIGN_UPGRADEABLE_ADMIN_BRIDGE` must be updated to `FOREIGN_BRIDGE_OWNER`, `FOREIGN_VALIDATORS_OWNER`, `FOREIGN_UPGRADEABLE_ADMIN`

After that the contracts `v2.2.0` should work for both set of e2e tests.

oracle-e2e/deploy.js
rzadp: Done.

ui-e2e/scripts/deploy.js
rzadp: Done.

Dockerfile
rzadp: Seems like wrong direction to me because:
- we need this Dockerfile in root dir because otherwise we wouldn't be able to access `yarn.lock` and other sub-dirs that are subject to e2e testing
Screenshot 2019-05-22 at 10 24 57
Not sure if we can or want to fight against docker in this regard.
- goes against flat repository structure
- not sure how `yarn workspaces` would work with `package.json` located two directory levels down (instead of just one down)

patitonar:
`oracle-e2e` and `ui-e2e` should be able to work with `2.3.0-rc0`. The reason is that there are some `.env` variables that need to be added/updated.

Some of these variables probably need to be added in some on the env files and some on others env files:
```
DEPLOYMENT_GAS_LIMIT_EXTRA=0.2
HOME_REWARDABLE=false
FOREIGN_REWARDABLE=false
VALIDATORS_REWARD_ACCOUNTS=0x0000000000000000000000000000000000000000
DEPLOY_REWARDABLE_TOKEN=false
ERC20_EXTENDED_BY_ERC677=false
BLOCK_REWARD_ADDRESS=0x0000000000000000000000000000000000000000
DPOS_STAKING_ADDRESS=0x0000000000000000000000000000000000000000
```

ui-e2e/envs/contracts.env
```diff
@@ -32,3 +32,5 @@ FOREIGN_GAS_PRICE=10000000000
 
 REQUIRED_NUMBER_OF_VALIDATORS=1
 VALIDATORS="0xaaB52d66283F7A1D5978bcFcB55721ACB467384b"
+
+DEPLOYMENT_GAS_LIMIT=5000000
```
akolotov: you already have this parameter in the line 4 but with a different value.

ui-e2e/envs/erc-contracts.env
```diff
@@ -33,3 +33,5 @@ ERC20_TOKEN_ADDRESS=0x3C665A31199694Bf723fD08844AD290207B5797f
 
 REQUIRED_NUMBER_OF_VALIDATORS=1
 VALIDATORS="0xaaB52d66283F7A1D5978bcFcB55721ACB467384b"
+
+DEPLOYMENT_GAS_LIMIT=5000000
```
akolotov: you already have this parameter in the line 4 but with a different value.

ui-e2e/envs/erc-native-contracts.env
```diff
@@ -34,3 +34,5 @@ ERC20_TOKEN_ADDRESS=0x3C665A31199694Bf723fD08844AD290207B5797f
 
 REQUIRED_NUMBER_OF_VALIDATORS=1
 VALIDATORS="0xaaB52d66283F7A1D5978bcFcB55721ACB467384b"
+
+DEPLOYMENT_GAS_LIMIT=5000000
```
akolotov: you already have this parameter in the line 4 but with a different value.

Dockerfile
akolotov: Ok. Don't you mind to rename the docker file to `Dockerfile.e2e` in this case?

ui-e2e/envs/contracts.env
rzadp: Done.

ui-e2e/envs/erc-contracts.env
rzadp: Done.

akolotov:
@patitonar could you check one more time and approve the PR? thanks!

PR #61, opened 2019-05-22
Merge dotfiles

### LINKED ISSUE #20: Merge individual dotfiles into one
Dotfiles like `.eslintrc`, `.prettierrc` in individual sub-directories should probably be merged into one in root directory.

rzadp:
Merged / Moved to root dir:
- `.gitignore`
- `.editorconfig`
- `.nvmrc`
  - I chose 10.15 in order to follow `.nvmrc` in the contracts
- `.dockerignore`

Removed:
- `.nycrc` as nyc is not used

I think it closes #20


rzadp:
> Should `oracle/.dockerignore` be used by `oracle/Dockerfile` when the oracle image is built for production?

Good call, I have duplicated the `.dockerignore` so now we have 2 `Dockerfile's` and 2 corresponding `.dockerignore's`

oracle/.dockerignore
```diff
@@ -1,12 +1,6 @@
...
-.git*
-LICENSE
-README.md
-node_modules
+**/node_modules
+**/submodules
+**/Dockerfile
+**/.dockerignore
+**/docker-compose.yml
+.gitmodules
```
akolotov: Do you think the following files and directories are needed to be included in the oracle docker image?
```
docs
test
README.md
*.png
```

rzadp: Let's remove them in #66 , in which I will probably be able to use only one `.dockerfile`

rzadp: I have updated the `.dockerignore` on both PRs.

.dockerignore
```diff
@@ -1,7 +1,14 @@
...
-**/**/.git
+**/submodules
+**/Dockerfile
+**/.dockerignore
+**/docker-compose.yml
 .gitmodules
+**/.git
+**/docs
+**/*.md
+**/*.png
```
akolotov: Is it safe and will not exclude elements of UI in the tokenbridge UI?

rzadp: Sorry I didn't think of it, it probably will.

PR #71, opened 2019-05-24
Oracle & Monitor - Compiled ABI's

### LINKED ISSUE #21: Extract repeated submodules/poa-bridge-contracts into one in root dir

rzadp:
- Fixes #21

#### Work done:
- Removed ABI's checked into source control
- Sourcing ABI's from `contracts` for `oracle` and `monitor`
- Changed script name from `compile:contracts` to `import:contracts` in `oracle`, `monitor` and `ui`
- Modified `oracle/Dockerfile` to include compiled contracts
- Extracted project setup steps from `Dockerfile.e2e` into `package.json` scripts `install:deploy` and `prepare:contracts` so that the same scripts will be used outsite e2e tests.


monitor/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "start": "node index.js",
     "lint": "eslint . --ignore-path ../.eslintignore",
-    "lint:fix": "eslint . --fix"
+    "lint:fix": "eslint . --fix",
+    "import:contracts": "rm -r -f ./abis && cp -r ../contracts/build/contracts ./abis/"
```
akolotov: Does it make sense to copy the ABIs in three different sub-directories locating in the same directory? What if we change paths to ABIs in the sources as so they will work with `../contracts/build/contracts` directly instead of local copy of this directory?

package.json
```diff
@@ -25,6 +25,10 @@
     "oracle-e2e": "./oracle-e2e/run-tests.sh",
     "ui-e2e": "./ui-e2e/run-tests.sh",
     "clean": "rm -rf ./node_modules ./**/node_modules ./**/**/node_modules ./**/build",
+    "compile:contracts": "yarn workspace poa-parity-bridge-contracts run compile",
```
akolotov: Just for you information: I have created https://github.com/poanetwork/poa-bridge-contracts/issues/189 to change the legacy package name.

akolotov: Does it make sense to use `yarn run compile:contracts` in the lines 21 and 23?

rzadp: Good spot. Done.

monitor/package.json
rzadp: I agree - done.

rzadp: This required me to disable `node/no-unpublished-require` lint rule and override `ModuleScopePlugin` in `ui`

PR #75, opened 2019-05-27
Updates after deployment test

### LINKED ISSUE #43: Update default bridge repo in deployment
There is a `bridge_repo` configuration variable in `deployment/oracle` which defaults to `https://github.com/poanetwork/token-bridge`, which should probably be updated to this monorepo.

### LINKED ISSUE #69: Manual test of deployment
As a final exercise of stage 1 we should perform a manual deployment on Kovan/Sokol testnet following the README's, updating the instruction along

rzadp:
- Fixes #69
- Fixes #43 

For exercise #69 I did the following:

#### Local dev setup
- Started 2 parity testnets, rabbitmq and redis
- Followed the instructions to deploy `contracts` on the testnets
- Followed the instructions to launch the `oracle` (in `NATIVE_TO_ERC` mode)
- Followed the instructions to use the corresponding `monitor`
- Followed the instructions to run the corresponding`ui`
- Corrected/Updated the instructions as I was going through them

#### Remote production setup
- Launched Ubuntu 16.04 on Digital Ocean
- Used the `deployment` to deploy and launch the production Oracle
- Updated it to use the monorepo instead of `token-bridge` repo

#### Additionally
- Changed all `npm run` to `yarn` in order to not mix up two of these through the monorepo
- Removed `gh-pages` dependency and scripts in `ui`
- Removed duplicated overview in `ui`


akolotov:
@phahulin or @ArseniiPetrovich could you take a look at the changes related to the bridge deployment?

ui/README.md
```diff
@@ -121,20 +128,33 @@ If successful, you will see bridge processes run when you issue a command. For e
...
+
+```
+yarn install && yarn install:deploy && yarn compile:contracts
+```
+  * Go to UI sub-repository  
+```
+cd ui
+```
 
 _**Note**: The bridge UI configuration should be performed with an unprivileged Linux account or with the following flag `npm install --unsafe-perm`. [More information](https://docs.npmjs.com/misc/scripts#user)_
```
akolotov: This is not needed in this section. Similar note should be provided in the step 5 if it make sense for `yarn`.

deployment/oracle/README.md
```diff
@@ -97,6 +97,8 @@ To be used with the ansible-playbook command, for example:
 
 * `--private-key=<file_name>` - if private keyfile is required to connect to the ubuntu instance.
 
+* `--user=<username>` - connect as this username
+
 ## Bridge service commands
 
 The Bridge service is named `poabridge`. Use the default `SysVinit` commands to `start`, `stop`, `restart`, and `rebuild` the service and to check the `status` of the service.
```
andogro: Not sure what would be required @akolotov , but may want to rename the service to `tokenbridge` to retain consistency.

deployment/oracle/README.md
```diff
@@ -53,17 +53,17 @@ cp hosts.yml.example hosts.yml
...
-2.1 `compose_service_user` - specifies users to be created by playbooks. This user will be used to run POA bridge.
+2.1 `compose_service_user` - specifies users to be created by playbooks. This user will be used to run POA Bridge Oracle.
 
-2.2 `bridge_repo` contains address of token-bridge repository. The default value is  https://github.com/poanetwork/token-bridge.
+2.2 `bridge_repo` contains address of POA Bridge Oracle repository. The default value is  https://github.com/poanetwork/tokenbridge.
 
 2.3 `bridge_repo_branch` points to the specific branch or commit to use with the `bridge_repo`. If `bridge_repo_branch` is not specified, the default (`master`) branch is used.
 
-2.4 `bridge_path` set the path where token-bridge would be installed. By default it point to the home folder of `compose_service_user`
+2.4 `bridge_path` set the path where POA Bridge Oracle would be installed. By default it point to the home folder of `compose_service_user`
```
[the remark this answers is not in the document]
andogro: `bridge_path` sets the path where the Token Bridge Oracle will be installed. By default it points to the home folder of `compose_service_user`

ui/README.md
```diff
@@ -95,13 +89,26 @@ The following is an example setup using the POA Sokol testnet as the Home networ
     * When deployment is finished, check that the `bridgeDeploymentResults.json` file exists in the `poa-bridge-contracts/deploy` directory and includes the bridge contract addresses.  
 
 5. Install and run the POA Token Bridge Oracle.
```
andogro:
```suggestion
5. Install and run the Token Bridge Oracle.
```

oracle/README.md
```diff
@@ -57,8 +57,10 @@ For more information on the Redis/RabbitMQ requirements, see [#90](/../../issues
 
 1. [Deploy the bridge contracts](https://github.com/poanetwork/poa-bridge-contracts/blob/master/deploy/README.md)
 
-2. Open `bridgeDeploymentResults.json` or copy the JSON output generated by the bridge contract deployment process.
+2. Open `bridgeDeploymentResults.json` and copy the JSON output generated by the bridge contract deployment process.
```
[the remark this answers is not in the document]
rzadp: Gotcha. Changed back.

monitor/README.md
```diff
@@ -133,17 +138,25 @@ GAS_PRICE_FALLBACK=21
...
+```
+
+## Check unprocessed events
+```
 node checkWorker2.js
-# check stuck transfers called by transfer, not transferAndCall
-# only applicable for bridge-rust-v1-native-to-erc
+```
+
+## Check stuck transfers called by transfer, not transferAndCall
```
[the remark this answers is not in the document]
rzadp: Removed the readme part and the `checkWorker3.js` file itself.

rzadp:
Last commit achieves the following:
- Updated root readme initialization instructions
- Pointed to root instructions in `oracle`, `ui` and `monitor` instead of duplicating instructions
- Moved note about `--unsafe-perm` to root dir

ui/README.md
rzadp: Fixed, please see [here](https://github.com/poanetwork/tokenbridge/pull/75#issuecomment-496453969)

ui/README.md
```diff
@@ -121,20 +128,33 @@ If successful, you will see bridge processes run when you issue a command. For e
...
-    * Update submodules  
-`git submodule update --init --recursive` 
-    * Install dependencies  
-`yarn install`
-    * Go to UI sub-repository  
-`cd ui`
+  * Clone the repository
+
+```
+git clone --recursive https://github.com/poanetwork/tokenbridge && cd tokenbridge
```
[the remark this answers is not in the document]
rzadp: Please see [here](https://github.com/poanetwork/tokenbridge/pull/75#issuecomment-496453969)

monitor/README.md
```diff
@@ -121,7 +121,12 @@ Example of an API
 ```
 
 # How to run
-Create .env file (see `.env.example` for parameters reference)
+Create .env file:
```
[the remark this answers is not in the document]
rzadp: Fixed, please see [here](https://github.com/poanetwork/tokenbridge/pull/75#issuecomment-496453969)

deployment/oracle/README.md
rzadp: Looks like it is just a simple search-and-replace - should I do it? @akolotov

akolotov: let's create a new issue for this

phahulin:
I launched an aws instance, changed value for `bridge_repo_branch: deployment` in `deployment/oracle/roles/repo/defaults/main.yml` and run the playbook on `example.yml` bridge with remote logging enabled (`syslog_server_port` in `hosts.yml`).
Got the following error on `jumpbox : Launch container` step:
```
No handlers could be found for logger "compose.config.environment"
```
Not sure where it's coming from, do you have any ideas?

The whole output is here https://gist.github.com/phahulin/e0a2d120c88d8d13de62171f4e011674
(along with more readable `module_stdout` from the last step)

rzadp:
> Not sure where it's coming from, do you have any ideas?
Hmm I don't have a clue what might be causing this.

akolotov:
> > Not sure where it's coming from, do you have any ideas?
> 
> Hmm I don't have a clue what might be causing this.

@rzadp  does it means that the issue was not reproduced on your setup (`bridge_repo_branch` in `main.yml` is `deployment` and `syslog_server_port` is configured in `hosts.yml`)?

rzadp:
> > > Not sure where it's coming from, do you have any ideas?
> > 
> > 
> > Hmm I don't have a clue what might be causing this.
> 
> @rzadp does it means that the issue was not reproduced on your setup (`bridge_repo_branch` in `main.yml` is `deployment` and `syslog_server_port` is configured in `hosts.yml`)?

No, I mean that I also have this issue now and not sure how to approach it.
Previously I left `syslog_server_port` commented out.

akolotov:
My suggestion is to merge this PR and open another one to investigate and fix the issue with `syslog_server_port` configured. @phahulin what do you think?

monitor/README.md
```diff
@@ -121,7 +121,22 @@ Example of an API
 ```
```
akolotov: Just discovered that the information about `GET /stuckTransfers` is not applicable anymore since related to call . Please remove it and the file `checkWorker3.js`. By doing this we could also close https://github.com/poanetwork/tokenbridge/issues/79

rzadp: Done.

akolotov: thanks

akolotov: please remove `checkWorker3.js` as well from the repo.

rzadp: It has already been removed.

akolotov: Oh. Great! It seems that `stuckTransfers.js` is not used anymore so it could be also removed.

deployment/oracle/README.md
akolotov: Please put the link to the issue here.

phahulin:
> My suggestion is to merge this PR and open another one to investigate and fix the issue with syslog_server_port configured. @phahulin what do you think?

Let's try to investigate it a little bit more. I think it should be a minor issue, but without it we don't have remote logging which is annoying

monitor/README.md
rzadp: Done - removed `stuckTransfers.js`

akolotov:
> > My suggestion is to merge this PR and open another one to investigate and fix the issue with syslog_server_port configured. @phahulin what do you think?
> 
> Let's try to investigate it a little bit more. I think it should be a minor issue, but without it we don't have remote logging which is annoying

@phahulin please correct me if I wrong. In fact whether we configure `syslog_server_port` or the following parts are triggered anyway since we have default value for `syslog_server_port` in [deployment/oracle/roles/logging/defaults/main.yml](https://github.com/poanetwork/tokenbridge/blob/deployment/deployment/oracle/roles/logging/defaults/main.yml)

```
- name: Set the log configuration file to send container logs to remote server
  template:
    src: 35-remote-logging.conf.j2
    dest: /etc/rsyslog.d/35-remote-logging.conf
    owner: root
    group: root
    mode: 0644
  when: syslog_server_port is defined
```
It means [deployment/oracle/roles/logging/templates/35-remote-logging.conf.j2](https://github.com/poanetwork/tokenbridge/blob/deployment/deployment/oracle/roles/logging/templates/35-remote-logging.conf.j2) is always parsed.

So, if we say that without `syslog_server_port` the issue we are discussing does not occur, but if it is configured, the error happens, does it mean that it could relate to the exact form of URL which is pointed out in `hosts.yml`?

akolotov:
Another thing I found is even if  the docker compose version is `1.22.0` in [the dependencies](https://github.com/poanetwork/tokenbridge/blob/48efc618e45ff337b2a9e7cfa157788653ea991d/deployment/oracle/roles/dependencies/defaults/main.yml), [the output](https://gist.github.com/phahulin/e0a2d120c88d8d13de62171f4e011674) reports about `"com.docker.compose.version": "1.24.0"`

akolotov:
Could it relate to that docker-compose cannot find `.env` file since we changed the path to the oracle?
https://github.com/poanetwork/tokenbridge/blob/5159e057ba80b357f6e2b34988054504ea4b179b/oracle/docker-compose.yml#L34-L38

ArseniiPetrovich:
Hi, guys! 

@akolotov You were nearly right - this is definitely connected with the changes we made to the path. If we check both `30-docker.conf.j2` and `35-remote-logging.conf.j2` we will see approximately the same logic, which is:
`if $programname startswith 'bridge_' then`.
It used to work until we moved the bridge to the oracle folder. According to the `docker-compose` name convention each container without explicit name will be started with the name like: `<project>_<service>_<index>_<slug>`, where `<project>` is folder where project is kept. 
I suggest either change the logic to `if $programname startswith 'oracle_bridge_' then` or name each container and... yeah, well, change the logic anyway.

However, this does not resolve the issue with `logger` bug, I will try to fix it now.

ArseniiPetrovich:
What's connected to the `logger` bug - I didn't find the roots of the issue, but discovered, that Ansible `docker_compose` module is not maintanied, so there is a couple of known bugs already, that is not fixed for a year.
So I suppose we can switch to using the `shell` module in the `jumpbox` role, at least until module bug will be fixed:
```
- name: Launch container
  shell: docker-compose up -d
  args:
    chdir: "{{ bridge_path }}/oracle"
```
Something like that should work.

rzadp:
I have:
- Updated the path in `30-docker.conf.j2` and `35-remote-logging.conf.j2` templates
- Used `shell` in place of the `docker_compose ` module
- Confirmed that docker logs appear in `/var/log/docker/*/docker.log`
Screenshot 2019-05-31 at 11 52 39

@ArseniiPetrovich @phahulin Could you please confirm that this is all that we needed?

Issue #57, opened 2019-05-22
TokenBridge UI in the docker environment


akolotov:
Consider to have an option to run the TokenBridge UI in the docker container in order to minimize requirements to the system where the UI could be deployed.

PR #82, opened 2019-05-31
CI performance impovements

### LINKED ISSUE #65: Duplicated actions in the run of e2e tests on CircleCI
As part of the initial integration to the monorepo both sets of e2e tests for the oracle (https://github.com/poanetwork/tokenbridge/pull/45) and for the UI (https://github.com/poanetwork/tokenbridge/pull/54) were added to CircleCI. 

From the CI process point of view both e2e test environment performs the following set of actions:
- pull parity docker image
- pull redis db docker image
- pull rabbitmq docker image
- pull node.js docker image
- prepare Linux on the oracle image to resolve dependencies
- resolve dependencies to run the oracle
- install dependencies to compile contracts
- compile contracts
- install dependencies to deploy contracts
- deploy contracts

It definitely increases time of the e2e tests and cause redundant load on the testing system.

Consider to optimize the process of running e2e testing on CircleCI in order to reduce number of duplicated actions.

rzadp:
#### The following has been done in order to improve CI performance:
- Introduced an `Initialize` job that initializes the project with dependencies and compiles the contracts
  - The result, saved in CI cache, is used by subsequent `build`, `lint`,  `test` and `cover` jobs
- Yarn's cache is saved in CI cache to use in every `Initialize` run
- Compiled contracts are saved in CI cache to use in every `Initialize` run
  - The cache is based on a key with commit hash of the `contracts`, so if we update `contracts` submodule then the cache will be invalidated
- Saving `yarn install:deploy` output in a CI artifact in order to get rid of large wall of text in CI logs coming from `node-gyp` that does not conform to `--silent` switch

#### E2E
- The e2e sub-projects cannot make use of the `Initialize` job because all of the initialization, installing dependencies, compiling contracts happens in the docker build system.
- Enabling `docker layer caching` allows to cache built docker layers in the e2e jobs run
  - Apparently the `docker layer caching` performance varies, not always the layers are taken from cache. I have included slower and quicker runs in the results

#### Notes
- From what I've read the `docker layer caching` feature might become a paid feature in the future
- I didn't find any more improvements, also I don't think that we can share pulling parity, redis etc. docker images between the e2e jobs. I'd say we close #65 but keep eye on further opportunities for improvement

#### Before (current master)
Screenshot 2019-05-31 at 15 36 43

#### After, slowest run
Screenshot 2019-05-31 at 15 37 22

#### After, quicker `oracle-e2e` run
Screenshot 2019-05-31 at 15 37 39

#### After, quicker `ui-e2e` run
Screenshot 2019-05-31 at 15 48 43

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,9 @@ jobs:
            keys:
              - yarn-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
       - run: yarn install --frozen-lockfile
-      - run: yarn install:deploy
+      - run: mkdir /tmp/artifacts && yarn install:deploy &> /tmp/artifacts/install_deploy.log
+      - store_artifacts:
+          path: /tmp/artifacts/install_deploy.log
       - run: yarn compile:contracts
       - save_cache:
            name: Save Yarn Package Cache
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
            name: Restore Yarn Package Cache
            keys:
              - yarn-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
-      - run: yarn install
+      - run: yarn install --frozen-lockfile
       - run: yarn install:deploy
       - run: yarn compile:contracts
       - save_cache:
@@ -59,8 +59,8 @@ jobs:
     steps:
         - checkout
         - run: git submodule update --init
-        - run: yarn
-        - setup_remote_docker
+        - setup_remote_docker:
+            docker_layer_caching: true
         - run: yarn run oracle-e2e
   ui-e2e:
     machine:
@@ -77,7 +77,7 @@ jobs:
       - run: wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
       - run: sudo dpkg -i google-chrome-stable_current_amd64.deb
       - run: git submodule update --init
-      - run: yarn
+      - run: yarn install --frozen-lockfile
       - run: yarn run ui-e2e
   cover:
     docker:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -59,30 +90,45 @@ jobs:
       - run: wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
       - run: sudo dpkg -i google-chrome-stable_current_amd64.deb
       - run: git submodule update --init
-      - run: yarn
+      - restore_cache:
```
akolotov: is it needed to restore cache in the `oracle-e2e` target as well?

rzadp: It's not, because:
- In `ui-e2e` we install dependencies in a CI step, so it's worth to cache yarn package cache
- In `oracle-e2e` we don't install any dependencies in a CI step (only in the context of a docker build)

The difference comes from the fact that `oracle-e2e` tests are run in a docker container, whereas the `ui-tests` selenium tests need to run in a [machine executor](https://circleci.com/docs/2.0/executor-types/#using-machine)

akolotov: Hmm... I see.

Does it mean that caching could be implemented when https://github.com/poanetwork/tokenbridge/issues/57 is addressed? Will it speed up CI checks?

rzadp: Not exactly because in `ui-e2e`, the `UI` already does run in a docker container (although using `Dockerfile.e2e` and not dedicated production Docker image like `oracle`).

What needs to be run in a machine executor is the headless chrome + selenium tests, which are accessing UI exposed from the docker container.

Having run the `ui-e2e` tests many times on my machine, I noticed that waiting for confirmations is taking a large portion of the time. Maybe we should create a separate issue to try and reduce required confirmations / speed up block mining?

patitonar: Right, the `ui` waits for 8 block confirmations even if the `oracle` processed the transactions after waiting 1 block as are configured on deployed contracts for `ui-e2e` tests. There is an issue created for this on the old bridge-ui repo https://github.com/poanetwork/bridge-ui/issues/119 to wait for the configured value on the contracts

akolotov: I will move the issues from the standalone repos on this week.

PR #83, opened 2019-05-31
 Add support v1 legacy contracts on monitor

patitonar:
Built on top of #81 

Add support for v1 bridge contracts so we can have one codebase of the monitor for all the deployed bridges.


monitor/index.js
```diff
@@ -83,6 +83,16 @@ app.get('/alerts', async (req, res, next) => {
   }
 })
 
+app.get('/stuckTransfers', async (req, res, next) => {
```
akolotov: does it make sense to provide access to this endpoint only in case of the V1 bridge?

patitonar: Thanks for the suggestion! Done https://github.com/poanetwork/tokenbridge/pull/83/commits/369b90a6d5866408b3c679ca0f527166b999342c

PR #86, opened 2019-06-04
Introducing E2E Commons

rzadp:
- Fixes #64

#### Work done

- Merged common `oracle-e2e` and `ui-e2e` `.env` files into one set
- Merged `docker-compose.yml` files into one
  - Renamed `bridge`, `bridge-erc20` and `bridge-erc20-native` docker services to `oracle`, `oracle-erc20` and `oracle-erc20-native`
- Extracted logic regarding spinning up e2e environment into one script
- Merged script regarding deploying e2e contracts into one script
  - Switched from JS script using shelljs to a shell script
- Merged constants from `oracle-e2e/config.json` and `ui-e2e/constants.json` into `e2e-commons/constants.json`
- Extracted helpers used to mine test blocks into e2e-commons
- Removed LICENCE from `oracle-e2e`

#### Ideas for further improvement
- #85

Issue #11, opened 2019-05-06
Add deployment-oracle testing in CI


rzadp:

rzadp:
@akolotov Should we close it having merged #113, and create separate for testing Monitor and UI deployment?

akolotov:
Yes, we can close it as soon as two new issues are created.

rzadp:
Added https://github.com/poanetwork/tokenbridge/issues/129 and https://github.com/poanetwork/tokenbridge/issues/130

PR #92, opened 2019-06-07
[WIP] [DISCUSSION] Deployment testing

rzadp:
### Purpose
The purpose of this PR is to have an early preview and discussion point about the work I started doing regarding deployment. Any comments/questions about the general direction would be appreciated.

I also have some problems and I would appreciate any tips/suggestions.

@akolotov @patitonar @phahulin @ArseniiPetrovich 

### Related issues
- #11
- #57 
- #9 

### Changes
This PR is branched out of `e2e-commons` #86 , which changes are not listed here.

- Moved `deployment/oracle/*` into `deployment/*`
    - The idea is to have only one main playbook `site.yml`
    - Have different host groups that would share roles and vars:
        - oracle
        - (to be added in #9) ui
        - (possibly) monitor
- Merged `deployment/README.md` and `deployment/oracle/README.md`
- #57 Introduced production `ui/Dockerfile`, use it in e2e instead of `Dockerfile.e2e`
- Introduced **dev deployment setup**

### Dev deployment setup
The idea is to have local docker containers running that would pretend to be oracle/ui/monitor host servers, and then run `site.yml` playbook against them. That way there is no need to spin up a VM in AWS or DigitalOcean in order to run the playbook against it.

This approach can be used to have deployment tested on CI. The first step in this direction I have in mind is to:
1. Run the playbooks against local docker containers
2. Assert that there were no errors in the playbook

#### What's been done so far

- Introduced `dev.yml` playbook
    - It spins up local docker container pretending to be oracle server
- Introduced `dev-hosts.yml` inventory - containing our local oracle server
- Successfuly run the `site.yml` against `dev-hosts.yml` inventory
    - I had to add some setup steps in `dev.yml` to make it work, like installing `rsyslog` in the oracle server - it probably needs to be moved to `site.yml`
    - I also had to modify the `site.yml` playbook a little bit to make it work here - like `docker service run` step

#### Problems
When trying out this approach, I run into very severe performance issues related to (I think) Docker-In-Docker situation.
We have a docker container that is an oracle-host, and then the `site.yml` playbook in role `jumbox` builds and run docker in it, which is a Docker-In-Docker.
The building process takes ridiculously long (both in CI on linux and on my macOS).
A simple Dockerfile step like `COPY yarn.lock .` can take tens of seconds...

I tried increasing memory and CPU cores dedicated to docker on my machine - monitoring with `docker stats` it looks like it was not runnning out of memory.

It's worth noting that we already have a Docker-in-Docker situation - if we take a look at `oracle-e2e` job in CI, we have a `circleci/node:10.15` docker container, and in this container we run `yarn run oracle-e2e` - which leads to building and running docker containers. And it's not slow like that.

So I guess that there must be something wrong. The slowness can be seen here on [CI](https://circleci.com/gh/poanetwork/tokenbridge/1449).
The `jumpbox : Build the container` task takes > 15 minutes, whereas locally (after `docker system prune -a` for fair comparision) it takes me 6 minutes to `cd oracle && docker-compose build`.


ArseniiPetrovich:
Hi, @rzadp!
Correct me if I wrong, but since you are using the `machine` build there is no `docker-in-docker`, or am I missing something? Also, I would suggest to try using the `setup_remote_docker` command as per https://circleci.com/docs/2.0/building-docker-images/. That might help.

rzadp:
> Hi, @rzadp!
> Correct me if I wrong, but since you are using the `machine` build there is no `docker-in-docker`, or am I missing something? Also, I would suggest to try using the `setup_remote_docker` command as per https://circleci.com/docs/2.0/building-docker-images/. That might help.

Thanks for your reply.
The `setup_remote_docker` is used only in `docker` executors and it doesn't apply for `machine` executors.

I'm using `machine` in this `deployment` job but docker-in-docker is there because:
1. First I run a docker container named `oracle-server` locally using `dev.yml` playbook
2. Then I run the `site.yml` playbook targeting this container
3. In task `jumpbox`, there is a `docker-compose`, which builds and runs an oracle container inside this oracle-server container (so docker-in-docker)

ArseniiPetrovich:
Got it, thank you.
But why can't we simply target the localhost then instead of creating the docker-container in terms of using the CircleCI? It will spin up the environment anyway.

rzadp:
> Got it, thank you.
> But why can't we simply target the localhost then instead of creating the docker-container in terms of using the CircleCI? It will spin up the environment anyway.

I will try that.
I think I still have to do the docker-container locally in order to not mess up local development device.

rzadp:
Closing for now.

PR #102, opened 2019-06-11
UI in docker

### LINKED ISSUE #57: TokenBridge UI in the docker environment

rzadp:
- Closes #57 
- Adds a dedicated `ui/Dockerfile`. The UI was already dockerized when running `ui-e2e` but was using `Dockerfile.e2e`
- Removes oracle, ui, monitor from the `Dockerfile.e2e`

### DIFF
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -6,7 +6,7 @@ COPY contracts/package.json ./contracts/
 COPY ui/package.json ./ui/
 COPY ui/lib/web3-eth/index.js ./ui/lib/web3-eth/index.js
 COPY yarn.lock .
-RUN yarn install --production
+RUN yarn install --production --frozen-lockfile
 
 COPY ./contracts ./contracts
 RUN yarn run compile:contracts
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -1,0 +1,18 @@
+FROM node:8
+
+WORKDIR /mono
+COPY package.json .
+COPY contracts/package.json ./contracts/
+COPY ui/package.json ./ui/
+COPY ui/lib/web3-eth/index.js ./ui/lib/web3-eth/index.js
+COPY yarn.lock .
+RUN yarn install --production
+
+COPY ./contracts ./contracts
+RUN yarn run compile:contracts
+RUN mv ./contracts/build ./ && rm -rf ./contracts/* ./contracts/.[!.]* && mv ./build ./contracts/
+
+COPY ./ui ./ui
+WORKDIR /mono/ui
+CMD echo "To start a UI application run:" \
+  "?"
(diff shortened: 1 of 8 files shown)


ui/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM node:8
```
akolotov: Please updates README to describe a procedure how to run the TokenBridge UI within the docker container.

rzadp: I have added a `ui/docker-compose.yml` and a step in the readme.

ui/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '2.4'
+services:
+  ui:
+    build:
+      context: ..
+      dockerfile: ui/Dockerfile
+    ports:
+    - "${PORT}:${PORT}"
+    command: 'true'
```
akolotov: my expectation is that `command` is not needed if `entrypoint` specified

ui/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '2.4'
+services:
+  ui:
+    build:
+      context: ..
+      dockerfile: ui/Dockerfile
+    ports:
+    - "${PORT}:${PORT}"
```
akolotov: seems that configuration of this parameter is not reflected in the readme file.

rzadp: The `${PORT}` variable is taken from `.env`, and there is a step about creating it from `.env.example` which contains a default PORT of 3001.

I have added a line to readme explaining that the UI will be served on `http://localhost:PORT`, because it is printed out by the react server, but it's not visible when running detached docker.

Will that be enough in this regard?

ui/docker-compose.yml
rzadp: Good spot. Looks like the same thing happens in our `oracle/docker-compose.yml`. Are you OK with creating a separate issue and PR for this in order to not mix up work and keep PRs small?

akolotov: We can create the issue to remove this line from existing `oracle/docker-compose.yml`. And seems that we introduce the new file in this PR, I would get rid of double work by committing correct version of the file. Do you see any objections?

ui/docker-compose.yml
akolotov: it is OK for me.

ui/docker-compose.yml
rzadp: Totally agree, I forgot that the file is being introduced and not already existing.
I created #106 for `oracle/docker-compose.yml`

akolotov: thanks!

PR #103, opened 2019-06-11
Improve sender retry strategy

### LINKED ISSUE #94: logs are being flooded with messages from the oracle sender in case of error
If by some reason the oracle sender cannot send a transaction it produces dozens of messages: https://gist.github.com/akolotov/465c35bd0656e0bb5ec74afceefe8542. It becomes very hard to investigate the issue since the logs are flooded with the same messages.

It is necessary to investigate why re-tries to send the message happens too fast and suggest a way to address the issue with logs.

patitonar:
See [this comment](https://github.com/poanetwork/tokenbridge/issues/94#issuecomment-500428242). When a tx fails, on rabbitmq a new queue is dynamically created (if not existed) with a message ttl calculated based on the retries attempts of that transaction. Once the message reaches the ttl, it is redirected to `deadLetterExchange` which will send the message back to the original queue that the sender is consuming. This new queue dynamically created will be autoremoved after a time of no activity.

Closes #94

### DIFF
--- a/oracle/src/sender.js
+++ b/oracle/src/sender.js
@@ -91,7 +91,7 @@ function updateNonce(nonce) {
   return redis.set(nonceKey, nonce)
 }
 
-async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
+async function main({ msg, ackMsg, nackMsg, channel, scheduleForRetry }) {
   try {
     if (redis.status !== 'ready') {
       nackMsg(msg)
@@ -167,7 +167,7 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
 
     if (failedTx.length) {
       logger.info(`Sending ${failedTx.length} Failed Tx to Queue`)
-      await sendToQueue(failedTx)
+      await scheduleForRetry(failedTx, msg.properties.headers['x-retries'])
     }
     ackMsg(msg)
     logger.debug(`Finished processing msg`)
--- a/oracle/src/services/amqpClient.js
+++ b/oracle/src/services/amqpClient.js
@@ -24,13 +24,17 @@ function connectWatcherToQueue({ queueName, cb }) {
 }
 
 function connectSenderToQueue({ queueName, cb }) {
+  const deadLetterExchange = `${queueName}-retry`
+
   const channelWrapper = connection.createChannel({
     json: true
   })
 
   channelWrapper.addSetup(channel => {
     return Promise.all([
+      channel.assertExchange(deadLetterExchange, 'fanout', { durable: true }),
       channel.assertQueue(queueName, { durable: true }),
+      channel.bindQueue(queueName, deadLetterExchange),
       channel.prefetch(1),
       channel.consume(queueName, msg =>
         cb({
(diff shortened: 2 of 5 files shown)


oracle/src/services/amqpClient.js
```diff
@@ -24,21 +24,39 @@ function connectWatcherToQueue({ queueName, cb }) {
...
       channel.consume(queueName, msg =>
         cb({
           msg,
           channel: channelWrapper,
           ackMsg: job => channelWrapper.ack(job),
           nackMsg: job => channelWrapper.nack(job, false, true),
-          sendToQueue: data => channelWrapper.sendToQueue(queueName, data, { persistent: true })
+          scheduleForRetry: async (data, msgRetries = 0) => {
+            const retries = msgRetries + 1
+            const delay = retries ** 2 * 1000
```
akolotov: we should limit this I think. as so it would not be too huge. E.g. 60 second between attempts is enough.

I would also suggest usage of Fibonacci sequence: 1, 2, 3, 5, 8, 13, 21, 34 or something like this. It will make short periods between first attempts (by hope that the situation is solved quickly) and provide reasonable delays if the situations was not fixed quickly.

oracle/src/services/amqpClient.js
```diff
@@ -24,21 +24,39 @@ function connectWatcherToQueue({ queueName, cb }) {
...
+            const retryQueue = `${queueName}-retry-${delay}`
+            await channel.assertQueue(retryQueue, {
+              durable: true,
+              deadLetterExchange,
+              messageTtl: delay,
+              expires: delay * 10
```
akolotov: what difference between `messageTtl` and `expires` is?

oracle/src/sender.js
```diff
@@ -167,7 +167,7 @@ async function main({ msg, ackMsg, nackMsg, sendToQueue, channel }) {
 
     if (failedTx.length) {
       logger.info(`Sending ${failedTx.length} Failed Tx to Queue`)
-      await sendToQueue(failedTx)
+      await scheduleForRetry(failedTx, msg.properties.headers['x-retries'])
```
akolotov: Does it make sense to have a test for this functionality?

oracle/src/services/amqpClient.js
patitonar: `messageTtl` is the time in which the message will be marked as dead and redirected to the dead letter Exchange.
`expires` is the time in which the queue will be deleted after disuse. where use means having consumers, being declared (asserted or checked, in this API).

For example with a `delay` value of 1 second:
- The queue will be created (`channel.assertQueue`)
- The message is pushed to the queue
- After one second message will expire and send to dead letter Exchange
- After 10 seconds of created/asserted the queue will be removed.

akolotov: Ok. thanks!

oracle/src/services/amqpClient.js
patitonar: Done https://github.com/poanetwork/tokenbridge/pull/103/commits/a466fe57dcfc9feb95acbf3421ece11f4980a5c6

oracle/src/sender.js
patitonar: Added unit tests https://github.com/poanetwork/tokenbridge/pull/103/commits/ae6692abcd20b45a4bdb0cb67928feb316e27746

oracle/test/test.env
```diff
@@ -1,2 +1,3 @@
 HOME_RPC_URL=http://example.com
 FOREIGN_RPC_URL=http://example.com
+QUEUE_URL=http://example.com
```
akolotov: Where is this URL used?

patitonar: When importing the method `getRetrySequence` on the unit test, the following line runs and try to connect to the amqp broker so I had to add the url to avoid getting the error on the connection.

https://github.com/poanetwork/tokenbridge/blob/3be812d55f819765b355c044a4a6f4023a9a6af8/oracle/src/services/amqpClient.js#L2

akolotov: Thanks for explanation

PR #109, opened 2019-06-14
Monitor in docker

### LINKED ISSUE #77: TokenBridge Monitor in the docker environment
Consider to have an option to run the TokenBridge Monitor in the docker container in order to minimize requirements to the system where the monitor could be deployed.

rzadp:
- Fixes #77 
- Merged `checkWorker` `checkWorker2` scripts into one `yarn check-all script`
- Updated `crontab.example`
- Removed contents of `.env.example` and `crontab.example` from readme
    - My reasoning is that since we point out that these files are available for reference, there is no need to add contents to readme file - plus we avoid situation where we update a file and forget about the readme

### DIFF
--- a/monitor/docker-compose.yml
+++ b/monitor/docker-compose.yml
@@ -1,0 +1,13 @@
+version: '2.4'
+services:
+  monitor:
+    build:
+      context: ..
+      dockerfile: monitor/Dockerfile
+    ports:
+    - "${PORT}:${PORT}"
+    env_file: ./.env
+    environment: 
+      - NODE_ENV=production
+    restart: unless-stopped
+    entrypoint: yarn start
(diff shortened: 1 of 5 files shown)


monitor/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '2.4'
+services:
+  monitor:
+    build:
+      context: ..
+      dockerfile: monitor/Dockerfile
+    ports:
+    - "${PORT}:${PORT}"
```
akolotov: We need to specify in README that the web interface is run on `PORT`. And extend the example of the configuration file with this parameter.

rzadp: Right. The `.env.example` already has this variable.
I updated the readme adding info about the web interface running on `PORT`

PR #112, opened 2019-06-17
Monitor - E2E

### LINKED ISSUE #12: Write e2e tests for TokenMonitor

rzadp:
Introduces end-to-end tests for the Monitor.

#### Coverage
The test cases cover basic integration and sanity checks:
- CheckWorker scripts work
- The endpoints are working and serving data on localhost

Closes #12 - I suggest we create a follow-up issue if we want to add more detailed test cases.

#### Bash script
The test cases were simple enough to fit into the `run-tests.sh` script. In case we add more detailed test cases it will probably make sense to add JavaScript tests like in `ui-e2e` and `oracle-e2e`.

#### Caveats for reference
- The executor of `monitor-e2e` job  is `machine`, because CircleCI `docker` executor, when using docker-in-docker, sets up a remote docker machine so `localhost:3003` does not simply work.
    [Docs](https://circleci.com/docs/2.0/building-docker-images/#accessing-services)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,4 +124,27 @@ workflows:
   version: 2
   tokenbridge:
     jobs:
+      - initialize
+      - initialize-root:
+          filters:
+            branches:
+              only: master
+      - build:
+          requires:
+            - initialize
+      - lint:
+          requires:
+            - initialize
+      - test:
+          requires:
+            - initialize
+      - cover:
+          requires:
+            - initialize
+          filters:
+            branches:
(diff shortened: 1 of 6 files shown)


rzadp:
@akolotov @patitonar Hi, did you have a chance to take a look?

rzadp:
> Looks good to me! As suggested, I think we should create a new issue for extending the test cases. Probably we should test that it works for the 3 bridge modes and check that some attributes are present on the data served

I created #115 for this.

Issue #115, opened 2019-06-27
Extend monitor E2E tests


rzadp:
- [x] Test that it works for the 3 bridge modes
- [ ] Check that some attributes are present on the data served
- [ ] Preferably rewrite to JS+mocha

rzadp:
@akolotov @patitonar Do you think that it is desirable to rewrite `monitor-e2e` from bash to JS+mocha, like `ui-e2e` and `oracle-e2e`?

My assessment is that it would pay off if we will keep extending the tests, but maybe it is not worth it if we will only add a couple of cases to check that some attributes are present on the data served, and that's it.

patitonar:
I think it would be nice to have the monitor e2e tests written in JS. For now is possible we have simple test cases but if eventually the monitor could be extended with some of the features described [here](https://github.com/poanetwork/tokenbridge/issues/32), I think it would be easier to extend the tests if they are in JS.

PR #116, opened 2019-07-01
Role rework plus deployment tests

rzadp:
### Role rework
Multiple roles for deploying Oracle has been merged into one Oracle role, with Common role as a dependency that will be re-used for future UI and Monitor roles.

### Deployment tests
Introduced deployment tests based on [Molecule](https://molecule.readthedocs.io).
The `site.yml` playbook is executed on a `ubuntu:16.04` docker container.
After that there are some assertion executed using [Testinfra](testinfra.readthedocs.io)


deployment/TESTING.md
```diff
@@ -0,0 +1,24 @@
+# POA Token Bridge / Deployment Testing
+
+The deployment playbooks are tested using [Molecule](https://molecule.readthedocs.io).
+
+## Prepare virtual python environment
+
+```
+command -v virtualenv || pip3 install virtualenv
```
akolotov: does it mean that to test deployment locally I need to install python3, virtualenv and other things listed in `requirements.txt`? Is it possible to containerize this by a docker file as so an adapter could test the deployment scripts without necessity to encumber the system with additional packages? It will definitely simplify the work process.

deployment/molecule/default/tests/test_oracle.py
```diff
@@ -0,0 +1,37 @@
...
+    ("poabridge"),
+    ("rsyslog")
+])
+def test_services(host, service):
+    assert host.service(service).is_enabled
+    assert host.service(service).is_running
+
+
+def test_remote_logging(host):
+    assert host.file('/etc/rsyslog.d/35-remote-logging.conf').exists
```
akolotov: We also need to check that logs are being written in proper place.

deployment/molecule/default/tests/test_oracle.py
```diff
@@ -0,0 +1,37 @@
...
+    ("oracle_bridge_senderhome_1"),
+    ("oracle_bridge_senderforeign_1"),
+])
+def test_docker_containers(host, name):
+    container = host.docker(name)
+    assert container.is_running
+
+
+@pytest.mark.parametrize("service", [
+    ("poabridge"),
```
akolotov: I am not sure that I understand which configuration (bridge contracts) will be used to for deployment testing. As part of testing we need to make sure that the tokenbridge oracle configured properly to work with the particular bridge contracts.

deployment/TESTING.md
rzadp: The packages listed in `requirements.txt` are not polluting the system but are installed in a folder much like `node_modules` in JS - that's thanks for `virtualenv`.

But yes, you still need `python3` and `virtualenv` installed.

I created #118 for this.

deployment/molecule/default/tests/test_oracle.py
rzadp: Unfortunately it cannot be tested - the reason is that we use `docker.sock` mounting:
```
volumes:
      - /var/run/docker.sock:/var/run/docker.sock
```

which is solving the docker-in-docker slowness, but also result in that the actual log files are on host machine running the molecule tests, instead of on `oracle-host`

deployment/molecule/default/tests/test_oracle.py
rzadp: Good question, I realized now that the tests are not ideal because:

The playbooks clone the monorepo on target-hosts using the `master` branch.

So on `branch-x`, if we modify playbooks, change oracle/ui/monitor code, update `contracts` submodule, then we test:
- Modified playbooks
- Unchanged oracle/ui/monitor from `master` branch
- Not-updated `contracts` submodule like in `master` branch

---

The desired behaviour, I presume, is that we test everything from `branch-x`.
I will try to make it happen in #119

rzadp:
@akolotov @ArseniiPetrovich Please take a look at my recent replies and follow-up issues I created.

If it's OK then I would love to merge the PR and do the follow-up issues in subsequent PRs in order to keep PRs small and short-living.

deployment/molecule/default/tests/test_oracle.py
akolotov: Good catch! My questions was actually whether the deployment e2e test works with contracts from `common-e2e`. The question appear since I do not see any configuration file for Ansible introduced by this PR allowing to work deployed contracts.

deployment/molecule/default/tests/test_oracle.py
akolotov: But will it be possible to test the logging if we specify another endpoint for the remote logs? Probably it will complicate the test logic but I am curious if we still have any way to test as much of the deployment functionality as we provide.

deployment/TESTING.md
akolotov: Thanks! I think we cannot finish the phase 2 without addressing #118.

deployment/molecule/default/tests/test_oracle.py
rzadp: I see. Currently the deployment tests do not use deployed contracts at all - just asserting that the playbook works and that the Oracle starts. It doesn't have E2E tests asserting that the Oracle works correctly (because we have that in `oracle-e2e`)

Seems like it could be beneficial to combine the molecule tests with e2e tests into ultimate e2e tests:
- Deploy the contracts using `e2e-commons`
- Deploy Oracle/UI/Monitor using Molecule, using the above contracts
- Run `oracle-e2e`, `ui-e2e` and `monitor-e2e` testing the above

But it doesn't seem quick nor easy.

akolotov: Exactly! Otherwise we are not testing the monorepo as the whole project by e2e tests. Instead the tests considers every component independently that does not prove that they will work together.

deployment/molecule/default/tests/test_oracle.py
rzadp: I'm not sure. I created #120 to investigate and try to achieve this.

akolotov: thanks

deployment/molecule/default/tests/test_oracle.py
rzadp: I created #121 for this.

akolotov: thanks! 
Currently my main worry is that you just run the poabridge service but since it was connected to any real contract we don't know if Ansible prepares proper configuration for the service. Do you think it should be covered in a separate issue as well?

rzadp: Yes, I would prefer to do it in a separate issue/PR, much like we did with #12 and #115 - keeping PRs rather small and short-lived

akolotov: OK.

akolotov:
Let's merge this PR as so others could be based on it.

PR #122, opened 2019-07-02
Deployment testing specific branch

### LINKED ISSUE #119: Deployment tests - branch specific
When running Molecule tests on `branch-x`, we test playbooks from `branch-x` with oracle/ui/monitor and contracts from `master` branch.

We should test everything from `branch-x`.

rzadp:
Closes #119 

The default `bridge_repo_branch` is `master`.
This PR overrides it with:
- `$CIRCLE_BRANCH` - on CI
- `$(git symbolic-ref --short HEAD)` - current local git branch

#### Caveat
If I'm developing on local branch `branch-x`, the playbook will not be able to `git clone` on target-host because `branch-x` does not exist on remote - that's why I added a section in readme about pushing.

### DIFF
--- a/deployment/TESTING.md
+++ b/deployment/TESTING.md
@@ -11,6 +11,22 @@ source venv/bin/activate
 pip install -r requirements.txt
 ```
 
+## Push remote branch
+
+The deployment playbooks are cloning the monorepository on target hosts, using your current local git branch name. If the branch does not exists on remote, you need to push it.
+
+```
+git push
+```
+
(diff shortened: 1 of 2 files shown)


rzadp:
> Does it work? :) I thought env variables in ansible are accessed via `lookup('env', ...)`

Not sure about ansible, but it does work in molecule which apparently is similar but different in some regards

rzadp:
https://molecule.readthedocs.io/en/stable/configuration.html#variable-substitution

PR #123, opened 2019-07-02
UI deployment with molecule tests

rzadp:
- Closes #9 
- Added `ui` to the deployment playbook
- Added molecule tests to default scenario (renamed to `example` scenario).
- Added `dai` and `wetc` scenarios to test the UI deployment with existing bridges
    - These I have manually tested by running `molecule converge` and inspecting the app served on localhost

PR #126, opened 2019-07-04
UI deployment

### LINKED ISSUE #9: Write playbooks for bridge-UI

rzadp:
- [x] service https://github.com/poanetwork/tokenbridge/pull/123#discussion_r300378188
- [x] hosts instructions https://github.com/poanetwork/tokenbridge/pull/123#discussion_r300373741
- Closes #9


rzadp:
@akolotov @ArseniiPetrovich @phahulin Please review again. The playbook is the same as in #123 , except for:
- Added some instructions to the readme regarding multiple hosts and standalone installations
- Added tokenbridge_ui service following Oracle's poabridge service
- Manual testing on digital ocean droplet required me to add `python3-pip` to dependencies after

deployment/roles/ui/tasks/servinstall.yml
```diff
@@ -0,0 +1,17 @@
+# This role creates a tokenbridge_ui service which is designed to manage docker-compose ui deployment.
+# /etc/init.d/tokenbridge_ui start, status, stop, restart - does what the services usually do in such cases.
+# /etc/init.d/tokenbridge_ui rebuild - builds a new ui deployment from scratch.
+---
+- name: "Set the service"
+  template:
+    src: tokenbridge_ui.j2
+    dest: "/etc/init.d/tokenbridge_ui"
```
akolotov: although there is not strict naming convention for names of the linux services I did a following experiment:
```shell
$ ls -1 /etc/init.d/ | grep '_' | wc -l
0
$ ls -1 /etc/init.d/ | grep '-' | wc -l
20
```
So, as you can see there is no services containing the underscore in the name. Do you have any objections to change the name as so it contain the hyphen instead of the underscore?

deployment/roles/ui/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+- include_tasks: pre_config.yml
+- include_tasks: jumpbox.yml
+- include_tasks: servinstall.yml
```
akolotov: does it make sense to have a task `logging.yml` especially for the standalone UI deployment?

deployment/CONFIGURATION.md
```diff
@@ -42,7 +64,7 @@ cp hosts.yml.example hosts.yml
 | VALIDATOR_ADDRESS_PRIVATE_KEY: `"<private_key>"` | The private key for the specified validator address. |
 | syslog_server_port: `"<protocol>://<ip>:<port>"` | Optional port specification for bridge logs. This value will be provided by an administrator if required.  |
 
-`hosts.yml` can contain multiple hosts and bridge configurations (groups) at once.
+`hosts.yml` can contain multiple bridge configurations at once.
```
akolotov: The line 78 in the next section refers to `Configuration parameters` of `oracle/README.md` to get more info about values that can be applied as the deployment configuration parameters. 
First, the link is broken.
The second now `configuration parameters` are the mix of the oracle and UI parameters. Moreover some parameters will be applied to both components and some are component specific. I think the section below needs to be updated to bring more light which parameters should be configured for which component as well as which parameters could be skipped in the case of standalone deployment.

phahulin: Also, this section lacks the actual command to launch ansible :)
```
ansible-playbook -i hosts.yml site.yml
```

rzadp:
@phahulin 
Please review latest changes.
I am using `python3` instead of `/usr/bin/python3` because for example on macOS it's in `/usr/local/bin/python3`. Do you think it's OK?

It also resolves https://github.com/poanetwork/tokenbridge/pull/126#discussion_r300490660

deployment/CONFIGURATION.md
rzadp: As per discussion on a call with @akolotov , created https://github.com/poanetwork/tokenbridge/issues/127 to have more clear instructions and consistent parameters.

@phahulin The command is in the `EXECUTION.md` file.

deployment/roles/ui/tasks/servinstall.yml
rzadp: Please review https://github.com/poanetwork/tokenbridge/pull/126/commits/c11012ba3935ab147954d91863898abeba546bcc
I have also modified documentation.

rzadp:
@phahulin I'm having a hard time reasoning about the differences in dependencies between running the playbook on real DigitalOcean droplet and running tests with molecule.
For example, after changing the `ansible_python_interpreter` the playbook worked on the droplet but the tests failed because `setuptools` were missing..

Anyway please review recent changes - I decided to go with explicit `python3` and `pip3`, I think it's more clear. The playbook works for me on the droplet and in molecule.

deployment/roles/ui/tasks/main.yml
rzadp: The PR is losing on clarity - I created https://github.com/poanetwork/tokenbridge/issues/128 to do this in a subsequent PR if that's OK with you.

phahulin: IMO it should be done in this PR, currently docker logs will keep accumulating eating up disk space.

akolotov:
> Deployment succeeded and I transferred 1 dai back and forth over the bridge +1
As far as I can tell, UI looks correct, @akolotov ping me on slack if you want to see the deployed version for yourself.

Thanks. No, your test is good for me.

akolotov:
@phahulin 
> But I think logging should be added in the scope of this PR because we already had a problem with docker logs in production.

I agree with you. But if the new PR related to logging will be addressed just after this PR, will it be OK for you?

deployment/roles/ui/tasks/main.yml
akolotov: OK. But please address it next

akolotov:
@rzadp I am OK with the changes then. Let's merge the PR and focus on the logging then.

PR #131, opened 2019-07-05
UI logging

### LINKED ISSUE #128: Add remote logging to UI deployment playbook
- Similarly as Oracle's `logging.yml`
- Reuse the `logging.yml` tasks if possible
- Make sure it does not interfere if we run Oracle+UI on one host

rzadp:
- Added remote logging for UI
- Changed order of tasks to setup logging first before starting components
- Changed `repo` task to allow deployment of multiple components
- Closes #128


rzadp:
I have hit a brick wall when trying to add remote logging for UI.
I'm following how it's done in Oracle but I don't quite understand these `.j2` templates as I have little knowledge on administering linux.

My idea is to add `31-docker.conf.j2` and `36-remote-logging.conf.j2` for UI, not worrying about code repetition for now. 

What happens is that seems like in this conditional:
```
if $programname startswith 'ui_' then \
  ?DockerLogFileName
else
  /var/log/docker/no_tag/docker.log
```

The `$programname startswith 'ui_'` condition is not true as only `no_tag` is created.

@phahulin @ArseniiPetrovich Could you shed a little light on this and help me?

ArseniiPetrovich:
Hi, @rzadp!
Let me help you a bit with understanding how syslog is working.
Each message it is receiving usually have the following format:
```
<PRI> TIMESTAMP HOST TAG MSG
```
The main thing we are interested in is a `TAG` which represents (or should represent) the actual application name, and it is always alphanumeric.
For the each log that is coming to the host Syslog server parse it in accordance with the list of `.conf` files. 
The `$programname` variable allows us to check the `TAG` of the message, which would be the application name. If looks like that the issue is in fact that your application name doesn't start from `ui_`. You can easily check it - simply start your application and track the syslog in the another terminal. You will see the correct name of the application you should check for.

Do not hesitate to contact me if you need more help with that or smth is still unclear ;)

rzadp:
@ArseniiPetrovich Thanks for your reply! It definitely helped!

Turns out the issue was that the remote logging setup task was executed after the `ui` app already started and did not produce any additional logs on its own. I have swapped the order of tasks to setup logging first.

Another change is to not force downloading the repository, otherwise Oracle's logging setup in `docker-compose.yml` is overwritten when downloading repo in UI role.

rzadp:
I think I have finished the work except one little detail.

For some reason, logs in `/var/log/docker/no_tag/docker.log` appear twice when deploying Oracle+UI on one host.

Screenshot 2019-07-08 at 17 03 40

I understand that they are logged there when `$programname startswith 'docker'`. I have extracted this condition from the j2 templates into separate `34-docker-remote-logging.conf.j2`, so that it's the same for both Oracle and UI, and overwritten when deploying one after another.

So we end up with exactly one `34-docker-remote-logging.conf` file in `/etc/rsyslog.d`, but still there are duplicates in `/var/log/docker/no_tag/docker.log`.

@ArseniiPetrovich Would you have any advice on how to investigate this as I don't know how to approach this?

ArseniiPetrovich:
Well, that behavior sounds super-logical to me, because 34-docker-remote-logging.conf is designed for logs redirection and not related to the local logic. However, you do have the `30-docker` and `31-ui-docker` conf files. Both of them are processed during the logs handling and both have the following line:
```
if $programname startswith '<whatever>' then \
  <LogFile>
else
  /var/log/docker/no_tag/docker.log
```
Since in both cases logs doesn't start from the matching string, they are redirected to the `no_tag` folder. And this is made twice because, well, two files.

Hope this helps,
Arsenii

rzadp:
Thanks for your help @ArseniiPetrovich !

I have fixed this and also extracted common templates+tasks to the `common` role.

There is one more thing - If my understanding is correct, the logging before my changes was set to include ALL logs from the whole system in `no_tag/docker.log` except the Oracle logs.
Was that a mistake? I have made the changes to include only `if $programname startswith 'docker'`.

If my reasoning is correct then I think the PR is ready for review.

deployment/molecule/default/tests/test_all.py
```diff
@@ -18,3 +18,8 @@ def test_docker_group(host):
 def test_user(host):
     assert host.user('poadocker').exists
     assert 'docker' in host.user('poadocker').groups
+
+
+def test_logging(host):
+    assert host.file('/etc/rsyslog.d/30-docker.conf').exists
```
phahulin: Is it possible to add a check (for this file and other log files) that permissions are `0644`?

deployment/roles/common/tasks/logging.yml
```diff
@@ -0,0 +1,29 @@
...
+    mode: 0644
+
+- name: Set the log configuration file to send docker logs to remote server
+  template:
+    src: 35-docker-remote-logging.conf.j2
+    dest: /etc/rsyslog.d/35-docker-remote-logging.conf
+    owner: root
+    group: root
+    mode: 0644
+  when: syslog_server_port is defined
```
phahulin: I think that commented out `syslog_server_port` option should also be added to `hosts.yml.example` in `ui` section as well

rzadp: Done.

deployment/molecule/default/tests/test_all.py
rzadp: Yes - Done.

The available assertions we can run are [here](https://testinfra.readthedocs.io/en/latest/modules.html#file).

Additionally if there are no assertions that we need, we can always use `host.run()` to run arbitrary shell command and check it's return code.

PR #135, opened 2019-07-05
Use mintedTotallyByBridge for bridge-ui statistics

### LINKED ISSUE #110: mintedTotally is used in UI instead of mintedTotallyByBridge
Currently `mintedTotally` is used to get the value of coins minted by `erc-to-native` bridge. But in common case it is not correct since it will shown value of coins minted by all bridges if more than one bridge is able to transfer tokens to the chain.

https://github.com/poanetwork/tokenbridge/blob/26ff22ac7ecc02027f7bcc2057f77b27f94a5c10/ui/src/stores/HomeStore.js#L256

The BlockReward contract provides ability to get number of coins minted by a specific bridge by calling `mintedTotallyByBridge`.

patitonar:
Closes #110


ui/src/stores/utils/contract.js
```diff
@@ -53,8 +53,8 @@ export const getBalanceOf = async (contract, address) => {
   return fromDecimals(balance, decimals)
 }
 
-export const mintedTotally = async contract => {
```
akolotov: is it possible to reflect this in tests?

patitonar: We could add some unit tests for this method, mocking the returned value from `contract.methods.mintedTotallyByBridge(bridgeAddress).call()` and also unit tests for `HomeStore` that could make sure that `mintedTotallyByBridge` method is called from the store

akolotov: yes, do this please

patitonar: Done

akolotov: thanks

PR #137, opened 2019-07-05
Remove redis lock in oracle sender

### LINKED ISSUE #107: Remove mutex used in the sender worker to control nonce increment
During testing of the token bridge oracle in some cases (https://github.com/poanetwork/token-bridge/issues/118, the root cause was not identified) the following issue appear:
```
[1556816976277] ERROR (LockError): Exceeded 200 attempts to lock the resource "lock:home:nonce".
    LockError: Exceeded 200 attempts to lock the resource "lock:home:nonce".
        at C:\dev\bridge-oracle\node_modules\redlock\redlock.js:342:20
        at tryCatcher (C:\dev\bridge-oracle\node_modules\bluebird\js\release\util.js:16:23)
        at Promise.errorAdapter [as _rejectionHandler0] (C:\dev\bridge-oracle\node_modules\bluebird\js\release\nodeify.js:35:34)
        at Promise._settlePromise (C:\dev\bridge-oracle\node_modules\bluebird\js\release\promise.js:566:21)
        at Promise._settlePromise0 (C:\dev\bridge-oracle\node_modules\bluebird\js\release\promise.js:614:10)
        at Promise._settlePromises (C:\dev\bridge-oracle\node_modules\bluebird\js\release\promise.js:690:18)
        at _drainQueueStep (C:\dev\bridge-oracle\node_modules\bluebird\js\release\async.js:138:12)
        at _drainQueue (C:\dev\bridge-oracle\node_modules\bluebird\js\release\async.js:131:9)
        at Async._drainQueues (C:\dev\bridge-oracle\node_modules\bluebird\js\release\async.js:147:5)
        at Immediate.Async.drainQueues [as _onImmediate] (C:\dev\bridge-oracle\node_modules\bluebird\js\release\async.js:17:14)
        at runCallback (timers.js:705:18)
        at tryOnImmediate (timers.js:676:5)
        at processImmediate (timers.js:658:5)
```
The attempt to protect the critical section in the code happens here:
https://github.com/poanetwork/tokenbridge/blob/ca1cf8061579fa94230a9771e76c447b92940ad4/oracle/src/sender.js#L108

This lock was introduced initially to provide the ability to run several sender workers on the same oracle instance. During the token bridge usage such scenarios were not confirmed that's why it was decided to remove the lock at all: this will simplify the logic a bit and address the issue described above.

The protection by the mutex can be introduced later if it is necessary.

patitonar:
Closes #107


oracle/src/sender.js
```diff
@@ -30,7 +30,6 @@ if (process.argv.length < 3) {
 const config = require(path.join('../config/', process.argv[2]))
 
 const web3Instance = config.web3
-const nonceLock = `lock:${config.id}:nonce`
```
akolotov: Should we consider a special upgrade procedure if an existing oracle instance is upgraded?

patitonar: I don't think we need a special upgrade procedure. Now the sender won't check for the nonce lock key on redis, so even if the nonce was locked and the sender was stopped before the unlock, when the upgraded sender starts again, it won't be affected

akolotov: thanks

PR #139, opened 2019-07-10
Added Commons sub-repository

rzadp:
- Introduced `Commons` sub-directory
- Extracted some `abi`s, constants and utils functions (but not using them yet)
- Integrated linting and testing `Commons` into CI
- First step of #8


commons/utils.js
```diff
@@ -0,0 +1,60 @@
...
+    case '0xba4690f5':
+      return BRIDGE_MODES.ERC_TO_ERC
+    case '0x18762d46':
+      return BRIDGE_MODES.ERC_TO_NATIVE
+    default:
+      throw new Error(`Unrecognized bridge mode hash: '${bridgeModeHash}'`)
+  }
+}
+
+const decodeFeeManagerMode = managerModeHash => {
```
akolotov: Don't think that getting the fee manager mode is used anywhere besides the UI. @patitonar please confirm.
Does it really make sense to move it to common part of the repo?

rzadp: It's true that it is used only in UI, but on the other hand it feels like it should be in the same place as `decodeBridgeMode` since they're very similar. If you think otherwise then I'll remove it and keep this in UI.

patitonar: Yes, it is used only in UI, but I agree that it makes sense to keep this method together with other decoding methods, so I'm OK to moving it to common

akolotov: great! I am approving the PR

Issue #8, opened 2019-05-06
Introduce bridge-commons


rzadp:

akolotov:
Thanks to @patitonar - he has formulated the set of improvements that could increase the part of the code in `commons` part of the monorepo:
```
Extract to commons:
- fetchGasPrice methods are similar on Oracle, monitor and UI
- getTokenType used on UI, monitor and Oracle (initialChecks.js)
- getValidatorList, processValidatorsEvents and parseValidatorEvent used on UI and monitor

Fix
- Oracle should use ERC_TYPES from commons
- UI HomeStore should use HomeBridgeV1Abi from commons
```

rzadp:
@akolotov @akolotov 
I have a question regarding `fetchGasPrice`.

Digging into the code and trying to make it common, I noticed some differences:

### Fallback

If the call to gas price API fails, then:

- `oracle` falls back to `bridgeContract.methods.gasPrice().call()`
    - If it fails, it falls back to `x_GAS_PRICE_FALLBACK`
- `monitor` and `ui` fall back to `x_GAS_PRICE_FALLBACK`, without calling the contract

### Limits

- `oracle` limits the fetched gas price within `GAS_PRICE_BOUNDARIES`
- `monitor` and `ui` do not

Should we update `monitor` and `ui` or should we preserve the differences?


---

### Naming

On a side note, the fact that we fetch the gas price from "Gas Price Oracle" is confusing, because we have the Oracle component. Maybe we should rename it?

Maybe like so:
- `HOME_GAS_PRICE_ORACLE_URL` => `HOME_GAS_PRICE_URL`
- `fetchGasPriceFromOracle` => `fetchGasPriceFromUrl`

patitonar:
@rzadp Thanks for the analysis, I think we should preserve the differences regarding fallback and limits since it is the correct behavior. Only `oracle` is required to call `bridgeContract.methods.gasPrice().call()` as a first fallback value and also to apply the gas price boundaries.

Regarding the naming, I agree that changing the name will help to avoid confusions with the oracle component, so I'm OK with it. Let's wait for @akolotov thoughts on this

akolotov:
I agree that the current differences regarding fallback and limits are OK:
- getting the fallback price for UI differs from getting the fallback price for the oracle since in the first case the price could be manually adjusted by the end user anyway, in the second way it is better to provide automatic way to get the best price,
[...]

PR #142, opened 2019-07-10
Add support for oracle gas price values not expressed in gwei

### LINKED ISSUE #105: Support ethgasstation as gasprice oracle
Please add support for gasprice oracle provided by https://ethgasstation.info/
API endpoint is
https://ethgasstation.info/json/ethgasAPI.json

The problem with this API is that gasprice values are in 0.1 GWei instead of GWei, so they would have to be divided by 10 before passing to sender.

As @akolotov suggested it would be handy to handle this more generally and introduce a new env variable - numeric (floating point) coefficient that would be multiplied by the value returned from gasprice API.

patitonar:
Closes #105 

Env variables `HOME_GAS_PRICE_FACTOR` and `FOREIGN_GAS_PRICE_FACTOR` were added, also in Monitor some other env variables were added too to properly specify values for home and foreign sides.

Now other gas price oracles as `https://ethgasstation.info/json/ethgasAPI.json` are supported on Oracle, UI and Monitor. (For example using `*_FACTOR=0.1`)

The implementation to fetch the values from the oracle is similar in all the bridge components, it could be  extracted in the future once https://github.com/poanetwork/tokenbridge/issues/8 is completed.

@phahulin Please check that the new env variables are correctly set on `deployment` files


monitor/validators.js
```diff
@@ -89,19 +105,40 @@ async function main(bridgeMode) {
...
+  )
+  const foreignGasPriceGwei = Web3Utils.fromWei(foreignGasPrice.toString(), 'gwei')
+  const foreignTxCost = foreignGasPrice.mul(Web3Utils.toBN(FOREIGN_GAS_LIMIT))
+
   let validatorsMatch = true
   logger.debug('calling asyncForEach foreignValidators foreignVBalances')
   await asyncForEach(foreignValidators, async v => {
     const balance = await web3Foreign.eth.getBalance(v)
-    const leftTx = new Web3Utils.BN(balance).div(txCost).toString(10)
+    const leftTx = foreignTxCost.isZero()
```
akolotov: Is `foreignTxCost.isZero()` possible? Even if so should `leftTx` be zero in this case? Does it make sense to report a very big number or something like `infinite`?

oracle/README.md
```diff
@@ -180,10 +180,12 @@ where the _watcher_ could be one of:
 | `HOME_GAS_PRICE_SPEED_TYPE` | Assuming the gas price oracle responds with the following JSON structure: `{"fast": 20.0, "block_time": 12.834, "health": true, "standard": 6.0, "block_number": 6470469, "instant": 71.0, "slow": 1.889}`, this parameter specifies the desirable transaction speed. The speed type can be omitted when `HOME_GAS_PRICE_ORACLE_URL` is not used. | `instant` / `fast` / `standard` / `slow` |
 | `HOME_GAS_PRICE_FALLBACK` | The gas price (in Wei) that is used if both the oracle and the fall back gas price specified in the Home Bridge contract are not available. | integer |
 | `HOME_GAS_PRICE_UPDATE_INTERVAL` | An interval in milliseconds used to get the updated gas price value either from the oracle or from the Home Bridge contract. | integer |
+| `HOME_GAS_PRICE_FACTOR` | A value that will multiply the gas price of the oracle to convert it to gwei. If the oracle API returns gas prices in gwei then this can be set to `1`. | integer |
```
akolotov: Please extend the description to point out that this factor could be used to intentionally pay more gas than suggested by the oracle in order to guarantee the transaction verification in the moment of sudden growth of gas price. E.g. `1.25` or `1.5` could be still fine in this case.

monitor/validators.js
```diff
@@ -89,19 +105,40 @@ async function main(bridgeMode) {
...
   const foreignVBalances = {}
   const homeVBalances = {}
-  logger.debug('calling getGasPrices')
-  const gasPriceInGwei = await getGasPrices(GAS_PRICE_SPEED_TYPE)
-  const gasPrice = new Web3Utils.BN(Web3Utils.toWei(gasPriceInGwei.toString(10), 'gwei'))
-  const txCost = gasPrice.mul(new Web3Utils.BN(GAS_LIMIT))
+
+  logger.debug('calling home getGasPrices')
+  const homeGasPrice = await getGasPrices(
+    HOME_GAS_PRICE_ORACLE_URL,
```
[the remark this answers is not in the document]
patitonar: If fetching the data from the oracle fails, it will use the fallback gas price parameter. Added a special check to avoid calling `fetch` if url is not defined so there is no error on the logs. https://github.com/poanetwork/tokenbridge/pull/142/commits/b22a0a2a22de0524a47e5cad4c7aa134d021891c

monitor/validators.js
patitonar: You are right, I don't think zero is possible on foreign. Removed that check for foreign. On home side, if `homeTxCost` is `zero`, `leftTx` will be `999999`, I tried to use `Infinity` number, but it does not play well with `JSON.stringify` to write the results. https://github.com/poanetwork/tokenbridge/pull/142/commits/b22a0a2a22de0524a47e5cad4c7aa134d021891c

deployment/roles/oracle/templates/.env.j2
```diff
@@ -23,6 +23,7 @@ HOME_GAS_PRICE_FALLBACK={{ HOME_GAS_PRICE_FALLBACK }}
 {% if HOME_GAS_PRICE_UPDATE_INTERVAL | default('') != '' %}
 HOME_GAS_PRICE_UPDATE_INTERVAL={{ HOME_GAS_PRICE_UPDATE_INTERVAL }}
 {% endif %}
+HOME_GAS_PRICE_FACTOR={{ HOME_GAS_PRICE_FACTOR }}
```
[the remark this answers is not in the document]
patitonar: Thanks, fixed https://github.com/poanetwork/tokenbridge/pull/142/commits/bfe3b2a3c94354bcdfcd8d33b4a6a2f2d60d82a4

deployment/roles/ui/templates/.env.j2
```diff
@@ -22,11 +22,13 @@ REACT_APP_HOME_GAS_PRICE_SPEED_TYPE={{ HOME_GAS_PRICE_SPEED_TYPE }}
 {% endif %}
 REACT_APP_HOME_GAS_PRICE_FALLBACK={{ UI_HOME_GAS_PRICE_FALLBACK }}
 REACT_APP_HOME_GAS_PRICE_UPDATE_INTERVAL={{ HOME_GAS_PRICE_UPDATE_INTERVAL }}
+REACT_APP_HOME_GAS_PRICE_FACTOR={{ HOME_GAS_PRICE_FACTOR }}
```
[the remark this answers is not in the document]
patitonar: Done https://github.com/poanetwork/tokenbridge/pull/142/commits/ee416cb28e21fa55a1bf7c113e08f498231c720b

deployment/group_vars/dai.yml
```diff
@@ -22,12 +22,14 @@ FOREIGN_POLLING_INTERVAL: 5000
 HOME_GAS_PRICE_SPEED_TYPE: "standard"
 HOME_GAS_PRICE_FALLBACK: 0
 HOME_GAS_PRICE_UPDATE_INTERVAL: 600000
+HOME_GAS_PRICE_FACTOR: 1
```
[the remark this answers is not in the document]
patitonar: Done https://github.com/poanetwork/tokenbridge/pull/142/commits/7b97fc339db23a6d04324d57985dc4c73ae1b186

deployment/roles/monitor/templates/.env.j2
```diff
@@ -4,9 +4,16 @@ HOME_BRIDGE_ADDRESS={{ HOME_BRIDGE_ADDRESS }}
 FOREIGN_BRIDGE_ADDRESS={{ FOREIGN_BRIDGE_ADDRESS }}
 HOME_DEPLOYMENT_BLOCK={{ MONITOR_HOME_DEPLOYMENT_BLOCK }}
 FOREIGN_DEPLOYMENT_BLOCK={{ MONITOR_FOREIGN_DEPLOYMENT_BLOCK }}
-GAS_PRICE_SPEED_TYPE={{ HOME_GAS_PRICE_SPEED_TYPE }}
-GAS_LIMIT={{ MONITOR_GAS_LIMIT }}
-GAS_PRICE_FALLBACK={{ MONITOR_GAS_PRICE_FALLBACK }}
+HOME_GAS_LIMIT={{ MONITOR_HOME_GAS_LIMIT }}
+HOME_GAS_PRICE_ORACLE_URL={{ HOME_GAS_PRICE_ORACLE_URL }}
```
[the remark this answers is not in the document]
patitonar: Added the checks here https://github.com/poanetwork/tokenbridge/pull/142/commits/00134b24766c7bda7e7cfe50d1bb77361debc8f7

oracle/README.md
patitonar: Updated description, please check https://github.com/poanetwork/tokenbridge/pull/142/commits/f7bb0a869e9d18388a5ff701425bc6dd3b2f5ca0

deployment/group_vars/dai.yml
```diff
@@ -45,7 +46,9 @@ UI_FOREIGN_GAS_PRICE_FALLBACK: 10000000000
 MONITOR_PORT: 3003
 MONITOR_HOME_DEPLOYMENT_BLOCK: 759
 MONITOR_FOREIGN_DEPLOYMENT_BLOCK: 6478417
-MONITOR_GAS_LIMIT: 300000
-MONITOR_GAS_PRICE_FALLBACK: 21
+MONITOR_HOME_GAS_LIMIT: 300000
+MONITOR_FOREIGN_GAS_LIMIT: 300000
+MONITOR_HOME_GAS_PRICE_FALLBACK: 21
```
akolotov: `MONITOR_HOME_GAS_PRICE_FALLBACK` should be zero

deployment/group_vars/dai.yml
```diff
@@ -45,7 +46,9 @@ UI_FOREIGN_GAS_PRICE_FALLBACK: 10000000000
...
+MONITOR_FOREIGN_GAS_PRICE_FALLBACK: 21
```
akolotov: Should `MONITOR_FOREIGN_GAS_PRICE_FALLBACK` be equal `FOREIGN_GAS_PRICE_FALLBACK`?

deployment/group_vars/example.yml
```diff
@@ -45,7 +47,9 @@ UI_FOREIGN_GAS_PRICE_FALLBACK: 1000000000
 MONITOR_PORT: 3003
 MONITOR_HOME_DEPLOYMENT_BLOCK: 0
 MONITOR_FOREIGN_DEPLOYMENT_BLOCK: 0
-MONITOR_GAS_LIMIT: 300000
-MONITOR_GAS_PRICE_FALLBACK: 21
+MONITOR_HOME_GAS_LIMIT: 300000
+MONITOR_FOREIGN_GAS_LIMIT: 300000
+MONITOR_HOME_GAS_PRICE_FALLBACK: 21
```
akolotov: Should they be equal `HOME_GAS_PRICE_FALLBACK` and `FOREIGN_GAS_PRICE_FALLBACK`?

deployment/group_vars/wetc.yml
```diff
@@ -45,7 +47,9 @@ UI_FOREIGN_GAS_PRICE_FALLBACK: 10000000000
 MONITOR_PORT: 3003
 MONITOR_HOME_DEPLOYMENT_BLOCK: 7703292
 MONITOR_FOREIGN_DEPLOYMENT_BLOCK: 7412459
-MONITOR_GAS_LIMIT: 300000
-MONITOR_GAS_PRICE_FALLBACK: 21
+MONITOR_HOME_GAS_LIMIT: 300000
+MONITOR_FOREIGN_GAS_LIMIT: 300000
+MONITOR_HOME_GAS_PRICE_FALLBACK: 21
```
akolotov: Should they be equal `HOME_GAS_PRICE_FALLBACK` and `FOREIGN_GAS_PRICE_FALLBACK`?

monitor/.env.example
```diff
@@ -4,8 +4,18 @@ HOME_BRIDGE_ADDRESS=0xABb4C1399DcC28FBa3Beb76CAE2b50Be3e087353
...
 HOME_DEPLOYMENT_BLOCK=0
 FOREIGN_DEPLOYMENT_BLOCK=0
-GAS_PRICE_SPEED_TYPE=standard
-GAS_LIMIT=300000
-GAS_PRICE_FALLBACK=21
+
+HOME_GAS_LIMIT=300000
+HOME_GAS_PRICE_ORACLE_URL=https://gasprice.poa.network/
+HOME_GAS_PRICE_SPEED_TYPE=standard
+HOME_GAS_PRICE_FALLBACK=1000000000
```
akolotov: Does the monitor wait the gas price fallback value in gwei?

patitonar: Good catch! All parameters for gas price in UI, Oracle and contract deployment are expected to be in `wei`, monitor is the only one that expect the gas price fallback value in `gwei`. Does it makes sense to define it here in `wei` too for consistency and to avoid confusions? Actually the changes introduced in this PR for monitor expect the gas price fallback to be in `wei`. If we want to follow this way, we only will need to update the env vars in `deployment` to also be in `wei`

akolotov: > Does it makes sense to define it here in wei too for consistency and to avoid confusions?

I believe it could be done when functionality to work with gas price oracles is part of `commons`.

akolotov: > If we want to follow this way, we only will need to update the env vars in deployment to also be in wei

What minimum amount of changes needs to be done to address this issue within this PR?

patitonar: Well on this PR `HOME_GAS_PRICE_FALLBACK` and `FOREIGN_GAS_PRICE_FALLBACK` were introduced and expressed in `wei` for monitor and the code also expect them to be in `wei`. So the only pending changes would be to update `MONITOR_HOME_GAS_PRICE_FALLBACK` and
`MONITOR_FOREIGN_GAS_PRICE_FALLBACK` to be in `wei` instead of `gwei` in files from `deployment/group_vars/`

patitonar: Here I applied the changes, also related to other comments https://github.com/poanetwork/tokenbridge/pull/142/commits/7d940b33bd27c1f8d7c2d1a933a8c4b58725be89

akolotov: Ok. Now I see what you meant. Thanks!

PR #143, opened 2019-07-11
Monitor using Commons

rzadp:
- Next step of #8 
- Replaced all references to ABIs, constants, utilities with `Commons`


monitor/getBalances.js
```diff
@@ -14,16 +14,16 @@ const web3Home = new Web3(homeProvider)
 const foreignProvider = new Web3.providers.HttpProvider(FOREIGN_RPC_URL)
 const web3Foreign = new Web3(foreignProvider)
 
-const ERC20_ABI = require('../contracts/build/contracts/ERC20').abi
-const ERC677_ABI = require('../contracts/build/contracts/ERC677').abi
-const HOME_ERC_TO_ERC_ABI = require('../contracts/build/contracts/HomeBridgeErcToErc').abi
-const HOME_ERC_TO_NATIVE_ABI = require('../contracts/build/contracts/HomeBridgeErcToNative').abi
-const BLOCK_REWARD_ABI = require('../contracts/build/contracts/IBlockReward').abi
-const FOREIGN_ERC_TO_ERC_ABI = require('../contracts/build/contracts/ForeignBridgeErcToErc').abi
```
rzadp: This is replaces by `FOREIGN_ERC_TO_ERC_ABI ` from Commons which is a `ForeignBridgeErc677ToEr677`

rzadp:
> Try to use the monitor with these changes on existing POA, xDAI and WETC bridges.

@akolotov 
In addition to `monitor-e2e`, I manually tested with the `dai` and `wetc` configs we have in `deployment`.

It works fine for `dai`.
For `wetc`, there is a `legacy access request rate exceeded` error.

Screenshot 2019-07-12 at 11 21 07

Is is probably because of old Infura endpoint, which now enforces its v3 version?
I tested with monitor from `master` and the result is the same, so the error is not introduced in this PR.

akolotov:
> Is is probably because of old Infura endpoint, which now enforces its v3 version?

It is quite interesting. I will open an issue for this - we need to investigate the reason and provide a fix.

akolotov:
> I manually tested with the `dai` and `wetc` configs we have in `deployment`

what parameters do you need in order to test it on the POA Bridge?

rzadp:
> > I manually tested with the `dai` and `wetc` configs we have in `deployment`
> 
> what parameters do you need in order to test it on the POA-POA20 Bridge?

I think bridge addresses, RPC urls and deployment blocks for home and foreign should be enough (other parameters like `LEFT_TX_THRESHOLD` does not seem important I think for this little test and can be copied from dai for wetc)

```
HOME_RPC_URL: 
FOREIGN_RPC_URL: 
HOME_BRIDGE_ADDRESS: 
FOREIGN_BRIDGE_ADDRESS: 
MONITOR_HOME_DEPLOYMENT_BLOCK: 
MONITOR_FOREIGN_DEPLOYMENT_BLOCK: 
```

akolotov:
> I think bridge addresses, RPC urls and deployment blocks for home and foreign should be enough (other parameters like `LEFT_TX_THRESHOLD` does not seem important I think for this little test and can be copied from dai for wetc)

You can get URLs and contract addresses on https://bridge.poa.network/. The block numbers you can get from the block explorers by using the contract addresses.

rzadp:
> You can get URLs and contract addresses on https://bridge.poa.network/. The block numbers you can get from the block explorers by using the contract addresses.

I've done that and confirmed that it works.
I was expecting to fail with same error as `wetc` since it uses `https://mainnet.infura.io` as well, but no, it works fine.

akolotov:
> I've done that and confirmed that it works.

Thanks, so, it could be merged then.

PR #145, opened 2019-07-11
UI using Commons

rzadp:
- Next step of #8 
- Replaced all references to ABIs, constants, utilities with `Commons`


rzadp:
I'm having trouble with react/webpack on this one.
@patitonar Would you be able to give me a hand on this one?

The references to `Commons` is done in the same way as all the other `X using Commons` PRs which are working fine, but here there is some problem with react/webpack.

Seems like webpack is not including the required ABIs. I'm not super familiar with react/webpack and I'm not sure how to approach this.

Screenshot 2019-07-12 at 10 38 31

It is not caused by including file from external directory, because we already did include the abis from external `contracts` - the change is to include from external `commons` instead.

patitonar:
@rzadp the only difference between UI and the other components is that we use es6 `import` syntax with webpack and all the config it uses. 
I was able to make the imports works by updating abi imports on `commons/abi.js` from: 
```javascript
const path = '../contracts/build/contracts/'

const HOME_NATIVE_TO_ERC_ABI = require(`${path}HomeBridgeNativeToErc`).abi
```
to
```javascript
const HOME_NATIVE_TO_ERC_ABI = require('../contracts/build/contracts/HomeBridgeNativeToErc').abi
```

I don't know exactly which is the issue, it seems that webpack have some kind of conflict when trying to calculate the correct path for the abi

rzadp:
@patitonar Thanks for looking into this! This fixed the issue and the PR is ready for review.

Issue #129, opened 2019-07-05
Add deployment-UI testing in CI


rzadp:
Testing playbook introduced in #9

PR #147, opened 2019-07-11
UI deployment - Molecule tests

rzadp:
- Renamed existing scenario `default` to `oracle`
- Introduced `ui` scenario
- `prepare.yml` shared between both scenarios
- `test_all.py` cases shared between both scenarios
- Closes https://github.com/poanetwork/tokenbridge/issues/129


deployment/molecule/ui/tests/test_ui.py
```diff
@@ -0,0 +1,32 @@
+import os
+import pytest
```
akolotov: which test here does check that the UI is running - the server responds with a correct content?

rzadp: I've added two tests:
- Check that the server responds with html with proper title
- Checks that returned html does not contain `error` string (with case insensitivity)

akolotov: thanks!

akolotov:
Now we can merge it

Issue #121, opened 2019-07-02
Ultimate E2E tests


rzadp:
We should combine the molecule tests with e2e tests into ultimate e2e tests:

- Deploy the contracts using e2e-commons
- Deploy Oracle/UI/Monitor using Molecule, using the above contracts
- Run oracle-e2e, ui-e2e and monitor-e2e testing the above

akolotov:
I would prefer to have step-by-step (PR-by-PR) introduction of components in the ultimate test.
And first of all, we need to by deployment e2e testing we need to check whether Ansible prepares proper configuration for the service. For this poaservice (the oracle) needs to be run to communicate to a real contracts.

rzadp:
The 3 types of bridge, and 3 components form a table of 9 cases to take care of.
I created the table to track progress:

... | `native-to-erc` | `erc-to-erc` | `erc-to-native`
--- | --- | --- | ---
**`oracle-e2e`** | ✅ Done | TODO | TODO
**`ui-e2e`** | TODO | TODO | TODO
**`monitor-e2e`** | TODO | TODO | TODO

PR #149, opened 2019-07-15
Building containers in deployment

rzadp:
- Side-effect when working on #121 

The `jumpbox` task starts the services (Monitor, UI, Oracle) using `docker-compose up`.

Then given service is stopped, and started again in `servinstall`.
So it probably makes sense to only build the containers in `jumpbox`.

### DIFF
--- a/deployment/roles/monitor/tasks/jumpbox.yml
+++ b/deployment/roles/monitor/tasks/jumpbox.yml
@@ -1,5 +1,5 @@
 ---
-- name: Launch container
-  shell: docker-compose up -d
+- name: Build the containers
+  shell: docker-compose build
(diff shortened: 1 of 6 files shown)


rzadp:
For some reason, the `service` module does not start our services.. I added an explicit shell step to perform `service X start`

rzadp:
> Will the service be run automatically after deployment?

Yes, our molecule tests check that.

rzadp:
Another reason to not start the services in `jumpbox` step, is that after `jumpbox`, in Oracle, there is `post_config` that for example installs the validator private key and it does not make sense to run the validators before that happens.

PR #152, opened 2019-07-16
Remove default syslog server port

rzadp:
Having this default is against our documentation: 
Screenshot 2019-07-16 at 15 07 18

..which implies that if you do not set this parameter, the remote logging is not default. Right now it is not the case because we have this default and the playbook sets up remote logging always, if I am not mistaken.

### DIFF
--- a/deployment/roles/oracle/defaults/main.yml
+++ b/deployment/roles/oracle/defaults/main.yml
@@ -3,5 +3,4 @@ bridge_path: "/home/{{ compose_service_user }}/bridge"
 ALLOW_HTTP: no
 QUEUE_URL: amqp://rabbit
 REDIS_URL: redis://redis
-syslog_server_port: udp://127.0.0.1:514
 keyfile_path: "/root/.key"


deployment/roles/oracle/defaults/main.yml
```diff
@@ -3,5 +3,4 @@ bridge_path: "/home/{{ compose_service_user }}/bridge"
 ALLOW_HTTP: no
 QUEUE_URL: amqp://rabbit
 REDIS_URL: redis://redis
-syslog_server_port: udp://127.0.0.1:514
```
akolotov: I have discovered that it was already in https://github.com/poanetwork/deployment-bridge/blob/master/bridge-nodejs/roles/logging/defaults/main.yml. Could you confirm that was incorrect too?

rzadp: That is my understanding. @ArseniiPetrovich @phahulin What do you think?

phahulin: `udp://127.0.0.1:514` is syslog service on localhost.
I deployed a test node without `syslog_server_port` and it logs to local syslog anyway, so this change doesn't seem to break anything. But I'd like to ask @ArseniiPetrovich for his opinion too.

ArseniiPetrovich:
LGTM. We better should to remove this default.

However, I want to make a clarification. The remote logging is set only if `syslog_server_port` variable exists.
The logging to local syslog will be performed  regardless the variable configuration.
Please, check lines 21-36 of logging.yml tasks file in Oracle role.

If the default is set, both configs will be active. However, since, as Pavel mentioned, the default address is a localhost one, the end user will not see any difference.

PR #154, opened 2019-07-16
Ability to override constant RPC urls in oracle-e2e

rzadp:
Done in preparation for #121 , where the RPCs will be different from what we have in our constants.


rzadp:
> Can you provide more details how it should work? When these URLs will be provided through the environment variables? Why should they be different from the stored in constants?

Actually, when trying to explain this, I discovered that it is not really necessary..

Turns out I was trying to run the `oracle-e2e` tests without using `e2e-commons/docker-compose.yml` like it's done in `oracle-e2e/run-tests.sh`, which caused the URLs stored in our constants to not work.

So I think this can be closed - I will reopen if needed.

PR #155, opened 2019-07-16
First step of ultimate e2e testing

rzadp:
#### Purpose
The purpose of this PR is to add CI job and molecule scenario needed in #121 , in order to avoid huge PR.

#### Added
- Molecule scenario `ultimate-native-to-erc`
    - Used to spin up Oracle using our deployment playbook
- CI job, which:
    - Launches the `ultimate-native-to-erc` scenario
    - Launches `oracle-e2e` tests (left as a todo)

#### Comments
- The differences between Docker for Mac and linux Docker required me to add `DOCKER_LOCALHOST` variable
- The `QUEUE_URL`, `REDIS_URL`, `HOME_RPC_URL`, `FOREIGN_RPC_URL` have to be changed because the hostnames like `rabbit` or `parity1` do not work when using `network_mode: host`.
    - We have to use `network_mode: host` because otherwise the Oracle, deployed as a Docker-in-Docker, cannot access the Paritys
- The initial idea, as discussed on last call with @akolotov , was to change existing code in `e2e-commons/up.sh`. Now I think that it's better to add these tests side-by-side, potentially removing current e2e tests in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,34 +134,10 @@ jobs:
       docker_layer_caching: true
     steps:
       - checkout
-      - run:
-          name: Install node
-          command: |
-            set +e             
-            curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh | bash
-            export NVM_DIR="/opt/circleci/.nvm"
-            [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-            nvm install v10.15.3
-
-            echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
-            echo "[ -s \"$NVM_DIR/nvm.sh\" ] && . \"$NVM_DIR/nvm.sh\"" >> $BASH_ENV
-      - run: 
-          name: Install yarn
-          command: |
-            nvm use v10.15.3
-            curl -o- -L https://yarnpkg.com/install.sh | bash
-            echo 'export PATH=$HOME/.yarn/bin:$PATH' >> $BASH_ENV
-            source $BASH_ENV
-      - run: 
-          name: initialize
-          command: |
-            nvm use v10.15.3
-            yarn run initialize
+      - run: git submodule update --init
       - run:
           name: Prepare the infrastructure
-          command: |
-            nvm use v10.15.3
-            e2e-commons/up.sh deploy native-to-erc
+          command: e2e-commons/up.sh deploy native-to-erc
       - run: 'echo "TODO - Run the e2e tests on top of the infrastructure created by previous step"'
 workflows:
   version: 2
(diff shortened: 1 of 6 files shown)


rzadp:
@akolotov Please take a look, do you have any questions/comments?

e2e-commons/up.sh
```diff
@@ -5,8 +5,13 @@ set -e # exit when any command fails
 ./down.sh
 docker-compose build
 docker-compose up -d parity1 parity2 e2e
+export DOCKER_LOCALHOST="localhost"
 
 while [ "$1" != "" ]; do
+  if [ "$1" == "macos" ]; then
```
akolotov: When will this be used?

.circleci/config.yml
```diff
@@ -128,6 +128,17 @@ jobs:
...
+  ultimate-native-to-erc:
+    machine:
+      image: circleci/classic:latest
+      docker_layer_caching: true
+    steps:
+      - checkout
+      - run: git submodule update --init
+      - run:
+          name: Prepare the infrastructure
+          command: e2e-commons/up.sh deploy native-to-erc
```
akolotov: could you clarify what exactly will happen with this step? Will it setup parity nodes and deploy contracts?

rzadp: The `e2e-commons/up.sh deploy native-to-erc` command performs the following:

1. Spins up the parity nodes (`docker-compose up -d parity1 parity2 e2e`)
2. Deploys the contracts (`docker-compose run e2e e2e-commons/scripts/deploy.sh`)
3. Launches the Molecule scenario (`../deployment/molecule/molecule.sh ultimate-native-to-erc`)
    - The Molecule scenario spins up ubuntu container
    - Then executes the oracle deployment playbook on that container
4. (todo) Run the oracle-e2e tests

Let me know if anything needs more clarification. I'm also thinking about adding some kind of diagram to our readmes that would illustrate this, specially the parts with docker-in-docker.

e2e-commons/up.sh
rzadp: This is used by me on macOS (or any other developer with macOS wanting to run these tests).

It's not used on CI because on linux this workaround is not needed.
The workaround is recommended by the [docs](https://docs.docker.com/docker-for-mac/networking/)

akolotov: so, you run it manually in pipeline like this 'e2e-commons/up.sh macos deploy native-to-erc', right?

.circleci/config.yml
akolotov: Thanks for clarification! Yes, some document describing the molecule testing is definitely needed. Too many relations.

akolotov: When I reviewed the code I lost top and bottom of `up.sh` that's why I did not get where parity is started and arguments are shifted.

rzadp:
> so, you run it manually in pipeline like this 'e2e-commons/up.sh macos deploy native-to-erc', right?

Exactly.

e2e-commons/up.sh
rzadp: Exactly.

PR #158, opened 2019-07-18
Ultimate E2E tests

rzadp:
- Progresses #121 
    - Covers ultimate tests for native-to-erc type of bridge, and oracle-e2e tests
- Adds docs and diagrams about the ultimate e2e: [Ultimate](https://github.com/poanetwork/tokenbridge/blob/ultimate-e2e/e2e-commons/ULTIMATE.md)
- Removes the `macos` flag and `DOCKER_LOCALHOST` - thanks to combining the containers in an `ultimate` docker network, it is no longer needed


deployment/molecule/ultimate-native-to-erc/converge.yml
```diff
@@ -0,0 +1,3 @@
+---
+- import_playbook: ../../site.yml
+- import_playbook: ./overwrite-docker-compose.yml
```
rzadp: The modification of Oracle's `docker-compose.yml` file is necessary in order to add `ultimate` network - joining the containers with parity containers

.circleci/config.yml
```diff
@@ -138,7 +138,12 @@ jobs:
       - run:
           name: Prepare the infrastructure
           command: e2e-commons/up.sh deploy native-to-erc
-      - run: 'echo "TODO - Run the e2e tests on top of the infrastructure created by previous step"'
+      - run:
+          name: Wait for the Oracle to start
+          command: 'sleep 120'
```
rzadp: I did try to have a more smart way to wait - waiting until `"Required block confirmations obtained"` record appears in the logs or similar.

I had trouble making that work, with the docker-in-docker situation and all, and eventually gave up - at least for now.

akolotov: Will it possible to get data from RedisDB to get information about observed blocks? By doing this (or even by watching if it changes) we could confirm that the oracle is up and running

akolotov: @rzadp :arrow_up:

rzadp: Yes, it most likely is possible, I will try to make it happen.

rzadp: Done - monitoring redis value of `native-erc-collected-signatures:lastProcessedBlock`.

rzadp:
> Added a comment to investigate the ability to check if the oracle is running through getting valued from redisdb

This is done now.

PR #160, opened 2019-07-21
add optional shift decimals

branciard:
Using the bridge in ERC20 to NATIVE mode, I need an decimals shift.
As the the ERC20 used is an 9 decimals token on foreign chain and I bridge it to a 18 decimals native on home chain. 
I need to shift by 9 decimals on one way and -9 in the other way. 

Here a PR of this optional FOREIGN_TO_HOME_DECIMAL_SHIFT in bridge config for that.

if FOREIGN_TO_HOME_DECIMAL_SHIFT is not set or equal to 0. It will not impact the current code.

### DIFF
--- a/oracle/.env.example
+++ b/oracle/.env.example
@@ -31,6 +31,8 @@ FOREIGN_START_BLOCK=
 LOG_LEVEL=debug
 MAX_PROCESSING_TIME=20000
 
+FOREIGN_TO_HOME_DECIMAL_SHIFT=
+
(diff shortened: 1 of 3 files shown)


branciard:
Thank you for your feedback. I will reconsider this functionality on contracts side as explained.

akolotov:
@branciard I have created https://github.com/poanetwork/poa-bridge-contracts/issues/243 where I described my ideas how the functionality you are interesting in could be implemented. Let's use that issue for additional discussion if needed. Thanks

Issue #136, opened 2019-07-05
Add cron schedule to monitor deployment playbook


rzadp:
We should use cron and `CRON_SCHEDULE` to automatically run checks in the monitor.

PR #161, opened 2019-07-23
Add cron schedule to monitor deployment playbook

rzadp:
- Closes https://github.com/poanetwork/tokenbridge/issues/136
- Removed docker flag `disable-legacy-registry` which is [removed from Docker as of v18.03](https://docs.docker.com/engine/deprecated/)

#### Tested by performing:
- Deployed the monitor with `MONITOR_CRON_SCHEDULE` not-set
    - The tasks to set up cron are skipped, playbook finishes successfully
- Deployed the monitor with `MONITOR_CRON_SCHEDULE` set
    - The cron job runs on schedule and runs the checks
- Deployed the monitor with `MONITOR_CRON_SCHEDULE` set, then stopped the `tokenbridge-monitor` service
    - The cron job runs on schedule and skips checking


deployment/roles/monitor/tasks/cron.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Parse cron schedule
+  set_fact:
+    minute: "{{ MONITOR_CRON_SCHEDULE.split(' ')[0] }}"
```
phahulin: How does this work when `MONITOR_CRON_SCHEDULE` is not set?

rzadp: If `MONITOR_CRON_SCHEDULE` is not set or invalid, task fails.
The playbook continues because `ignore_errors: True`.
The `Add cron entry` is then skipped because `when: not set_fact_result.failed`

phahulin: I think we should always install cron for monitor (otherwise it just checks ones), and so add a default for `MONITOR_CRON_SCHEDULE`

rzadp: I agree. I made the changes in the playbook and the docs, and also renamed `MONITOR_CRON_SCHEDULE` to `monitor_cron_schedule` to match style.

PR #166, opened 2019-07-25
Remove deprecated config option

MarkusTeufelberger:
https://docs.docker.com/engine/deprecated/#--disable-legacy-registry-override-daemon-option <-- This option is deprecated AND removed and causes Docker to fail the restart when this file gets installed.


rzadp:
This makes sense to me - I also removed this in pending PRs #161 and #162

rzadp:
The change has been already merged in #162 so the PR appears empty.
I propose we keep this PR open until #167 is done so we can confirm passing tests.

PR #168, opened 2019-07-29
Ultimate e2e - Native to erc - UI

rzadp:
- Progresses #121 
- Introduces UI-e2e tests for native-to-erc type of bridge
- Using `machine` executor like in `ui-e2e` job because installing Chrome


.circleci/config.yml
```diff
@@ -160,6 +160,66 @@ jobs:
...
+          command: |
+            curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+            echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+            sudo apt-get update && sudo apt-get install yarn
+      - run:
+          name: Install npm packages
+          command: nvm use default; yarn install --frozen-lockfile
+      - run:
+          name: Prepare the infrastructure
+          command: e2e-commons/up.sh deploy native-to-erc
```
akolotov: Do we need the task `ultimate-native-to-erc-oracle` at all? My understanding is that ultimate e2e testing will run oracle+ui+monitor in one test environment as so transactions will be sent initially with UI, then oracle will relay assets and monitor will check that the state of the bridge contracts is OK

rzadp: I have given it a thought and I agree with you.

My initial thinking was to have 3 separate jobs:
1. Deploy `oracle` -> run `oracle-e2e`
2. Deploy `oracle`+`ui` -> run `ui-e2e`
3. Deploy `oracle`+`monitor` -> run `monitor-e2e`

But there is no need for that, plus with 3 types of bridge we would have 9 separate jobs..

I agree with you and I think we should have just one job for one type of bridge that would:
1. Deploy `oracle`+`ui`+`monitor` -> run `ui-e2e`+`oracle-e2e`+`monitor-e2e`

rzadp: I have made the changes to have one `ultimate-native-to-erc` job instead of `ultimate-native-to-erc-oracle` and `ultimate-native-to-erc-ui`, so please review again.

The last step to `Run the oracle-e2e tests` is maybe not necessary since we test the oracle indirectly through `ui-e2e`, I thought to leave it and possibly remove later if need be.

akolotov: Agree that the last step is not needed. The idea of the ultimate tests was to check that everything works properly after deployment through deployment playbooks. So, indirect testing of the bridge oracle during the UI testing is enough for these purposes. We have anyway the oracle tests in the `e2e-oracle` job, don't we?

rzadp: I have removed the last step.

Yes, we still have the `oracle-e2e` job.

akolotov: thanks!

PR #173, opened 2019-07-31
Skip ui-e2e on master

rzadp:
We have a failing CI badge because of flaky `ui-e2e`.

Screenshot 2019-07-31 at 10 13 16

I propose we don't run them on master to avoid these false negatives.
The `ui-e2e` will still be run on every PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,7 +185,11 @@ workflows:
             branches:
               only: master
       - oracle-e2e
-      - ui-e2e
+      - ui-e2e:
+          filters:
+            branches:
+              ignore:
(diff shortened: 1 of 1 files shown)


rzadp:
> I would leave the testing as is. Otherwise we are trying to hide the problem.

OK, I agree. So I will close the PR now, it can be re-opened if need be.

I don't know why are the `monitor-e2e` failing from time to time.

It fails during `yarn install` on step:
`[4/4] Building fresh packages...`

In theory, because of `yarn.lock`, we download and build exact same versions of packages, so I can't see any reason why they would fail from time to time.

PR #179, opened 2019-08-02
Common gasprice

rzadp:
- Progresses https://github.com/poanetwork/tokenbridge/issues/8
- Extracted common functions from `ui`, `oracle` and `monitor` to `commons`
- Moved some tests from `oracle` and `ui` to `commons`


oracle/src/services/gasPrice.js
```diff
@@ -33,53 +33,18 @@ const foreignBridge = new web3Foreign.eth.Contract(ForeignABI, FOREIGN_BRIDGE_AD
...
-      logger.error(`There was a problem getting the gas price from the contract. ${e.message}`)
-    }
-  }
-  return gasPrice
+const fetchGasPrice = async (speedType, factor, bridgeContract, oracleFetchFn) => {
+  const contractOptions = { logger }
+  const oracleOptions = { speedType, factor, limits: GAS_PRICE_BOUNDARIES, logger }
+  cachedGasPrice =
+    (await gasPriceFromContract(bridgeContract, contractOptions)) ||
+    (await gasPriceFromOracle(oracleFetchFn, oracleOptions)) ||
```
patitonar: I think the order is not correct, we should get gas price from the Oracle first, then from the contract, then fallback value.

So this should be re-defined as:
```javascript
cachedGasPrice =
  (await gasPriceFromOracle(oracleFetchFn, oracleOptions)) ||
  (await gasPriceFromContract(bridgeContract, contractOptions)) ||
  cachedGasPrice
```

rzadp: You're right, it's a mistake, the original order was to get from oracle/supplier first.

Updated.

PR #181, opened 2019-08-06
Common Validator utils

### LINKED ISSUE #8: Introduce bridge-commons

rzadp:
- Closes #8 

#### Moved to commons:
- `parseValidatorEvent` and `processValidatorsEvents`
    - with tests
- `getValidatorList`
- `getPastEvents` and `getDeployedAtBlock`
    - moved as well because they are called from `getValidatorList`


ui/src/stores/utils/contract.js
```diff
@@ -62,70 +62,7 @@ export const totalBurntCoins = async contract => {
...
-    event.raw.topics &&
-    event.raw.topics[0] === '0xe1434e25d6611e0db941968fdc97811c982ac1602e951637d206f5fdda9dd8f1'
-  ) {
-    const rawAddress = event.raw.data === '0x' ? event.raw.topics[1] : event.raw.data
-    const address = '0x' + rawAddress.slice(26)
-    event.event = 'ValidatorRemoved'
-    event.returnValues.validator = address
-  }
-}
+export const getValidatorList = async (address, eth) => commonGetValidatorList(address, eth, { logger: console })
```
rzadp: The `getValidatorList` and `getPastEvents ` functions are left here (instead of removing it and importing from `commons` everywhere), because it is mocked here:

https://github.com/poanetwork/tokenbridge/blob/055a444faea09797bbe9b4dc1a29a2b514e9cd23/ui/src/stores/__tests__/HomeStore.test.js#L28-L36

And I failed to mock importing these from `commons` in `HomeStore` and `ForeignStore`

commons/utils.js
```diff
@@ -66,6 +67,116 @@ const getUnit = bridgeMode => {
...
+    return validators
+  }
+
+  options.logger && options.logger.debug && options.logger.debug('getting validatorsEvents')
+
+  const deployedAtBlock = await tryCall(validatorsContract.methods.deployedAtBlock(), 0)
+  const fromBlock = options.fromBlock || Number(deployedAtBlock) || 0
+  const toBlock = options.toBlock || 'latest'
+
+  const validatorsEvents = await getPastEvents({
```
patitonar: This `getPastEvents` call should be updated to match the signature of the method

```javascript
const getPastEvents = async (contract, { event, fromBlock, toBlock, options }) => {
```
For example:
```javascript
  const validatorsEvents = await getPastEvents(new eth.Contract([], address), {
    event: 'allEvents',
    fromBlock,
    toBlock,
    options: {}
  })
```

rzadp: Right. Updated.

commons/utils.js
```diff
@@ -66,6 +67,115 @@ const getUnit = bridgeMode => {
...
+  }
+}
+
+const getDeployedAtBlock = async contract => tryCall(contract.methods.deployedAtBlock(), 0)
+
+const getPastEvents = async (contract, { event, fromBlock, toBlock, options }) => {
+  let events
+  try {
+    events = await contract.getPastEvents(event || 'allEvents', {
+      options: options || {},
```
patitonar: `options` is not a valid parameter for `getPastEvents`. The correct way will be to add content of `options` object directly. See example [here](https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html#id41). One posible fix could be:
```javascript
const getPastEvents = async (contract, { event, fromBlock, toBlock, options = {} }) => {
  let events
  try {
    events = await contract.getPastEvents(event || 'allEvents', {
      ...options,
      fromBlock,
      toBlock: toBlock || 'latest'
    })
```

commons/utils.js
```diff
@@ -66,6 +67,115 @@ const getUnit = bridgeMode => {
...
+      fromBlock,
+      toBlock: toBlock || 'latest'
+    })
+  } catch (e) {
+    if (e.message.includes('query returned more than 1000 results') && Number.isInteger(toBlock)) {
```
patitonar:
```suggestion
    if (e.message.includes('query returned more than 10000 results')) {
```

While testing this with the monitor, I found that Infura error for max result was updated to `10000`, see [medium post](https://blog.infura.io/infura-june-2019-update-9528c388a3b4)  (FYI @akolotov ).

Also, the check `Number.isInteger(toBlock)` was not present in the previous implementation of this code. Is it really necessary? From the code and manual testing the monitor,  it looks that `toBlock` is a BigNumber and not an integer and is not getting inside this `if` when the max result error is present.

You can use env variables for xDai bridge to test this
```
HOME_RPC_URL=https://dai.poa.network
FOREIGN_RPC_URL=https://mainnet.infura.io/v3/...
HOME_BRIDGE_ADDRESS=0x7301CFA0e1756B71869E93d4e4Dca5c7d0eb0AA6
FOREIGN_BRIDGE_ADDRESS=0x4aa42145Aa6Ebf72e164C9bBC74fbD3788045016
HOME_DEPLOYMENT_BLOCK=759
FOREIGN_DEPLOYMENT_BLOCK=6478417
```

akolotov: > While testing this with the monitor, I found that Infura error for max result was updated to 10000, see medium post (FYI @akolotov ).

Very nice! Thanks!

commons/utils.js
rzadp: Right, my mistake. Updated.

commons/utils.js
rzadp: I changed the statement to
`e.message.includes('query returned more than')`
which should continue working even if the exact number is changed again by infura.

As for the check for `Number.isInteger(toBlock)`, the idea was to prevent from getting inside this `if` in a situation where `toBlock == 'latest'`, because the calculations for middleBlock would not work.
I corrected this, and also removed repetition by specifying default parameters in the method's signature.

patitonar: > I changed the statement to
e.message.includes('query returned more than')
which should continue working even if the exact number is changed again by infura.

Thanks! That makes sense

> As for the check for Number.isInteger(toBlock), the idea was to prevent from getting inside this if in a situation where toBlock == 'latest', because the calculations for middleBlock would not work.
I corrected this, and also removed repetition by specifying default parameters in the method's signature.

Thanks for the explanation

PR #186, opened 2019-08-12
Ultimate tests for ERC-TO-ERC

rzadp:
Similarly as other ultimate tests.
Closes https://github.com/poanetwork/tokenbridge/issues/121


ui-e2e/test.js
```diff
@@ -131,6 +131,7 @@ test.describe('e2e-test for bridge.poa, version 1.5.0', async function() {
 
   test.describe('ERC TO ERC', async () => {
     test.it('User is able to open main page of bridge-ui ', async () => {
+      await foreignAccount.setMetaMaskNetwork()
```
rzadp: It was missing when running only ERC-TO-ERC, without running NATIVE-TO-ERC portion before that.

akolotov: Should we have the same in `erc-to-native` section bellow?

rzadp: In `erc-to-native` section it already exists: https://github.com/poanetwork/tokenbridge/blob/9172ad98101d72b23d3ded9a8dbe0bcb119976a5/ui-e2e/test.js#L236

rzadp: I think that this case illustrates that there is some room for improvement in the `ui-e2e` tests. The preconditions for tests like `await foreignAccount.setMetaMaskNetwork() ` are unclearly hidden in other tests, and the tests rely on sequential execution - I think we could create an issue to refactor the `ui-e2e` a little bit, using `describe` and `before` to make them more clear.

akolotov: yes, please create the issue with detailed description what is expected to do under it.

rzadp: https://github.com/poanetwork/tokenbridge/issues/188

PR #187, opened 2019-08-12
Monitor E2E rewrite

### LINKED ISSUE #115: Extend monitor E2E tests

rzadp:
- Rewrite `monitor-e2e` tests to JS+Mocha
- Add tests for checking that some attributes are present on the data served
- Closes #115


monitor-e2e/test/common.js
```diff
@@ -0,0 +1,87 @@
...
+        it('should contain withdrawals', () => assert(data.home.withdrawals === 0))
+      })
+
+      describe('foreign', async () => {
+        it('should contain deposits:', () => assert(data.foreign.deposits === 0))
+        it('should contain withdrawals', () => assert(data.foreign.withdrawals === 0))
+      })
+
+      describe('general', async () => {
+        it('should contain balanceDiff', () => assert(data.balanceDiff === 0))
```
akolotov: it also needed to have a test to check that the monitor catches situations when the bridge is unbalanced since a deposit or withdrawal operation is stuck.

monitor-e2e/test/common.js
```diff
@@ -0,0 +1,87 @@
...
+    describe('validators', async () => {
+      let data
+
+      before(async () => {
+        ;({ data } = await axios.get(`${type.baseUrl}/validators`))
+      })
+
+      it('home', () => assert(typeof data.home.validators === 'object'))
+      it('foreign', () => assert(typeof data.foreign.validators === 'object'))
+      it('requiredSignaturesMatch', () => assert(data.requiredSignaturesMatch))
```
akolotov: it make sense to develop a test to check that monitor handles situations when different bridge sides have different numbers for required signatures or different validators set.

monitor-e2e/test/common.js
rzadp: This was not a small change, so I created a separate PR targeting this PR, so it can be reviewed separately. Please take a look: https://github.com/poanetwork/tokenbridge/pull/191

monitor-e2e/test/common.js
rzadp: The `REQUIRED_NUMBER_OF_VALIDATORS` and `VALIDATORS` are not set separately for different sides of the bridge, when deploying.

Seems like, in order to test different values, we would have to deploy the bridge twice with different values, and then take Home bridge from one pair and take Foreign from another pair.

This seems like overhead for just this one test case. Do we still want to do it?
In the mean time, I have extended these tests a little bit in a PR into this PR: https://github.com/poanetwork/tokenbridge/pull/192

akolotov: You can change the set of validators later if you call `addValidator` on one side of the bridge.

https://github.com/poanetwork/poa-bridge-contracts/blob/048f51e0e8e7ce4213b727c0982aadab6c994db9/contracts/upgradeable_contracts/BridgeValidators.sol#L45

rzadp:
@akolotov After #191 and #192 this PR needs final review.

PR #191, opened 2019-08-21
[PR-into-PR] Monitor E2E rewrite - balance

rzadp:
https://github.com/poanetwork/tokenbridge/pull/187#discussion_r313583326

Extending the https://github.com/poanetwork/tokenbridge/pull/187 with a test to check that the monitor catches situations when the bridge is unbalanced since a deposit or withdrawal operation is stuck.

The balance difference is achieved by sending a transfer to one side of the bridge, and the bridge remains unbalances because the Oracle is not running in `monitor-e2e`.

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,7 +1,5 @@
 FROM node:10
 
-RUN curl -sSL https://get.docker.com/ | sh
-
 WORKDIR /mono
 COPY package.json .
 COPY oracle-e2e/package.json ./oracle-e2e/
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,5 +1,7 @@
 FROM node:10
 
+RUN curl -sSL https://get.docker.com/ | sh
+
 WORKDIR /mono
 COPY package.json .
 COPY oracle-e2e/package.json ./oracle-e2e/
--- a/monitor-e2e/test/nativeToErc.js
+++ b/monitor-e2e/test/nativeToErc.js
@@ -48,7 +48,7 @@ describe.only('NATIVE TO ERC with changing state of contracts', () => {
     assert(data.balanceDiff === 0)
 
     console.log('starting check all')
-    shell.exec('docker exec e2e-commons_monitor_1 yarn check-all')
+    shell.exec('docker exec e2e-commons_monitor_1 yarn check-all || docker exec e2e_commons_monitor_1 yarn check-all || docker exec e2ecommons_monitor_1 yarn check-all') // The difference between macOS and linux docker
     console.log('check all completed')
 
     ;({ data } = await axios.get(`${baseUrl}`))
--- a/monitor-e2e/test/nativeToErc.js
+++ b/monitor-e2e/test/nativeToErc.js
@@ -1,8 +1,16 @@
 const assert = require('assert')
 const axios = require('axios')
-const { nativeToErcBridge } = require('../../e2e-commons/constants.json')
+const Web3 = require('web3')
+const shell = require('shelljs')
+const { nativeToErcBridge, user, homeRPC, foreignRPC } = require('../../e2e-commons/constants.json')
+const { generateNewBlock } = require('../../e2e-commons/utils')
 
 const baseUrl = nativeToErcBridge.monitor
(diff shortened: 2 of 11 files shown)


monitor-e2e/utils.js
```diff
@@ -0,0 +1,42 @@
+const shell = require('shelljs')
+const Web3 = require('web3')
+const { ERC677_BRIDGE_TOKEN_ABI } = require('../commons')
+
+const checkAll = () => {
+  // the different prefixes come from macOS/linux docker differences and/or different docker-compose versions
+  ;['e2e-commons_', 'e2e_commons_', 'e2ecommons_'].forEach(prefix =>
+    ['monitor_1', 'monitor-erc20_1', 'monitor-erc20-native_1'].forEach(container =>
+      shell.exec(`docker exec ${prefix}${container} yarn check-all`)
```
rzadp: I need to run `docker exec e2e-commons_monitor_1` on macOS, and `docker exec e2ecommons_monitor_1` on CI.
I'm not exactly sure if the difference comes from macOS/linux difference, or docker-compose/docker version difference.
The `forEach` is supposed to try all combinations.

Dockerfile.e2e
```diff
@@ -1,5 +1,7 @@
 FROM node:10
 
+RUN curl -sSL https://get.docker.com/ | sh
```
rzadp: Needed for `checkAll` in monitor-e2e

monitor-e2e/test/nativeToErc.js
```diff
@@ -14,3 +15,19 @@ describe('NATIVE TO ERC', () => {
...
+  let data
+
+  before(async () => {
+    ;({ data } = await axios.get(`${baseUrl}`))
+    assert(data.balanceDiff === 0)
+  })
+
+  it('should change balanceDiff', async () => {
+    await sendEther(homeRPC.URL, user, nativeToErcBridge.home)
+    checkAll()
```
akolotov: could you clarify the logic of the test more? I don't understand it.

rzadp: 1. The precondition is that `balanceDiff === 0`
2. `sendEther` is used to send ether to only one side of the bridge
3. New data is fetched from the monitor
    - `checkAll()` so that the monitor runs it's `checkWorker` scripts
    - `axios.get` so that we get new data from the monitor
4. Expected result is that `balanceDiff !== 0`

akolotov: OK. Understood. This is due to the fact that our monitor currently is not re-generating data on periodical basis. In actual deployment we use `cron` for this purposes.
But supporting docker-in-docker environment just to work this limitation around is overhead for me.
I see several possible options that could help us:
1. Implement a new entry point in https://github.com/poanetwork/tokenbridge/blob/more-monitor-e2e-tests/monitor/index.js as so it will re-generate the data when receives the GET request. For security reasons this entry point could be available only in the testing environment. In this case the GET request can be sent after the `sendEther` call. 
2. Implement a new target for testing environment that will call a script that will re-generate the data every 2-4 seconds. So, instead of calling `yarn check-and-start`
https://github.com/poanetwork/tokenbridge/blob/1f611704414e5d40c7c71050a34b263d991b743c/e2e-commons/docker-compose.yml#L242
    it can call this script. 
3. Similarly to `wait-for-monitor.sh` in https://github.com/poanetwork/tokenbridge/blob/more-monitor-e2e-tests/monitor-e2e/run-tests.sh we can introduce a script that will re-generate the data on periodical manner in background. 

What do you think? I see that usage docker-in-docker approach is more straightforward but necessity to install docker only for these needs makes it unviable for me. Especially due to the issue you described here: https://github.com/poanetwork/tokenbridge/pull/191/files#r316196312

rzadp: I agree, and every option you proposed makes sense to me.
At first sight option `3.` seems most natural to me.
I will look into it more deeply and try to implement it.

rzadp:
@akolotov I have removed the changes for installing docker in `Dockerfile.e2e`, and I have implemented the re-generation of monitor data using:
- `periodically-check-all.sh` script
- `waitUntil` in tests

Please review again.

akolotov:
> Please review again.

It is OK for me now

PR #192, opened 2019-08-21
[PR-into-PR] Monitor E2E rewrite - validator

rzadp:
https://github.com/poanetwork/tokenbridge/pull/187#discussion_r313584109

Extending the https://github.com/poanetwork/tokenbridge/pull/187 with a little more explicit testing of the validators

### DIFF
--- a/monitor-e2e/test/common.js
+++ b/monitor-e2e/test/common.js
@@ -1,6 +1,6 @@
 const assert = require('assert')
 const axios = require('axios')
-const { nativeToErcBridge, ercToErcBridge, ercToNativeBridge } = require('../../e2e-commons/constants.json')
+const { nativeToErcBridge, ercToErcBridge, ercToNativeBridge, validator } = require('../../e2e-commons/constants.json')
 
 const types = [
   { description: 'NATIVE TO ERC', baseUrl: nativeToErcBridge.monitor },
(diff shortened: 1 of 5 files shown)


monitor-e2e/test/common.js
```diff
@@ -43,9 +43,21 @@ types.forEach(type => {
...
+
+      it('foreign', () => {
+        assert(typeof data.foreign.validators === 'object')
+        assert(data.foreign.validators[validator.address].balance > 0)
+        assert(data.foreign.validators[validator.address].leftTx > 0)
+        assert(data.foreign.validators[validator.address].gasPrice > 0)
+      })
+
+      it('requiredSignaturesMatch', () => assert(data.requiredSignaturesMatch, 1))
       it('validatorsMatch', () => assert(data.validatorsMatch))
```
akolotov: Consider to change the number of validators on one side of the bridge by invoking `addValidator` from the validators contract. The monitor should have `false` in `validatorsMatch` when the transaction is included in the block.

rzadp: Right, will do.

rzadp: I have done the changes in similar fashion as #191 

1. The precondition is that `validatorsMatch === true`
2. Then the `addValidator` is executed on only one side of the bridge
3. After `checkAll` and `axios.get`, the expected result is that `validatorsMatch === false`

akolotov: Thanks!

Dockerfile.e2e
```diff
@@ -1,5 +1,7 @@
 FROM node:10
 
+RUN curl -sSL https://get.docker.com/ | sh
```
akolotov: Let's wait for final decision in https://github.com/poanetwork/tokenbridge/pull/191. I provided there few options how we can avoid docker-in-docker environment.

rzadp:
@akolotov This has been updated after #191 and is ready again for review.

Issue #167, opened 2019-07-26
Tests do not handle forked Pull Requests


rzadp:
The first PR https://github.com/poanetwork/tokenbridge/pull/166 to the monorepo from a forked repository revealed that the deployment tests do not handle such kind of PRs.

The reason is that the deployment playbooks use `git pull`, and in order to test the deployment together with changes to codebase on a given branch, we pull "current branch".
But the "current branch" does on exist on our monorepo, it only exists on a forked repo - so `git pull` fails.

rzadp:
At the moment I don't have a good idea on how to handle this.
I guess one way would be to pull "current repo", in addition to "current branch"

@ArseniiPetrovich @phahulin @akolotov Do you have some idea on how to handle it?

rzadp:
I can think of another way to approach this - change the playbooks.

Instead of doing a `git pull` task, we would just use a task to `copy` the files we have locally.
Since locally we have just the files we want to test.
And also it would remove the need to `git push` first.

phahulin:
I think it's a good idea to just copy the files :+1:

akolotov:
correct. but we need to make sure that 
1) all files are added to the index
2) only added files are copied
It could prevent cases when the local tests are performed well but failed when the changes are pushed to the remote repo

PR #194, opened 2019-08-22
Molecule tests for deploying repository

rzadp:
- First step of #167 
- Introduces Molecule scenario for copying the repository
    - Tests files/directories for existence
- No changes to the ansible playbooks at this step


.circleci/config.yml
```diff
@@ -180,6 +180,15 @@ jobs:
     steps:
       - checkout
       - run: deployment/molecule/molecule.sh monitor
+  deployment-repo:
+    executor: tokenbridge-orb/machine-with-docker-caching
+    steps:
+      - checkout
+      - run: git submodule update --init
+      - tokenbridge-orb/install-node
```
akolotov: Seems I need an extended description what exactly we are solving by these changes. Could you provide it?

rzadp: 1. `install-node` and `install-yarn` is needed to perform `yarn-install-cached-on-machine`
2. `yarn-install-cached-on-machine` is needed so that there exists `node_modules`
3. Existing `node_modules` is needed so that we can test, that the new way of deployment will not copy local `node_modules` to remote server

rzadp:
> Could you provide an extended plan or description that will show how exactly #167 is going to be addressed.

@akolotov Sure, here is my plan:

### The problem(s)

The problem with #167 is that the deployment cannot perform a `git pull` for a branch that isn't there, in the monorepo.

Another problem is that, if you create a branch locally and try to run Molecule tests, you need to push your branch (otherwise again, the branch does not exist).
It is described [here](https://github.com/poanetwork/tokenbridge/blob/master/deployment/molecule/TESTING.md#push-remote-branch).

### The solution

The solution is to just copy the local files to target remote server, instead of performing a `git pull` remotely. This makes sense because we are already checked out on a branch we want to deploy/test.

The change turned out to be simple, I have actually implemented it on a branch but did not include it in the PR, wanting to have tests in place first with separate review.
The change looks like [this](https://github.com/poanetwork/tokenbridge/blob/9f9c810ac7392111aeca36b948253f08be2815ca/deployment/roles/common/tasks/repo.yml), replacing [current repo.yml](https://github.com/poanetwork/tokenbridge/blob/master/deployment/roles/common/tasks/repo.yml)

### Tests

The Molecule tests introduced in this PR have 3 objectives:
- Verify that the `repo.yml` task runs without error
- Verify that what must to be copied, gets copied
    - In `test_existing.py`
    - It doesn't really make sense to write too much test cases here. The files will change over time, plus if something would be missing then it would be caught by ultimate tests. I just put some basic sanity checks, so that if the task gets completely broken in the future then these tests would show it before the ultimates even start.
- Verify that what mustn't be copied, does not
    - `test_non_existing.py`
    - This is to make sure we don't copy over to remote server local `node_modules`, or local `.git` folder.

akolotov:
Thanks @rzadp for the description.
Could you clarify the following: If we `repo.yml` how will the repo be copied on CircleCI side?

deployment/molecule/repo/tests/test_non_existing.py
```diff
@@ -0,0 +1,17 @@
+import os
+import pytest
+import testinfra.utils.ansible_runner
+
+testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
+    os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
+
+
+@pytest.mark.parametrize("path", [
+    ("/home/poadocker/bridge/node_modules"),
```
akolotov: What should be done as part of the new approach in order to this directories will be copied? I mean, if someone changes the behavior intentionally to copy these directories why will he/she stop to not remove these checks?

rzadp:
> Thanks @rzadp for the description.
> Could you clarify the following: If we `repo.yml` how will the repo be copied on CircleCI side?

Every job on CircleCI starts with `- checkout`, which checks out the repo with a given PR branch.

deployment/molecule/repo/tests/test_non_existing.py
rzadp: If someone intentionally wants to copy these directories then he will remove the checks.

I have in mind unintentional copying, when configuration on what to ignore when copying is unintentionally broken.

rzadp: For example right now I am fighting with copying configuration to ignore `.git` when copying, but it doesn't work for me for some reason. 
Despite setting `exclude`, the `.git` gets copied to remote.
https://github.com/poanetwork/tokenbridge/blob/21480f25603853d27abd8572116332e36f88aa81/deployment/roles/common/tasks/repo.yml#L11
If I get it to work, still I think it would be good if it was covered by tests.

akolotov:
> Every job on CircleCI starts with `- checkout`, which checks out the repo with a given PR branch.

Thanks

deployment/molecule/repo/tests/test_non_existing.py
akolotov: I see now

rzadp:
> Does it make sense to extend the test to check that files are only copied are in the list `git ls-tree -r HEAD --name-only`?

Yes, I think it would be very good. Should we do it in this PR or in a follow-up Issue?

akolotov:
> Yes, I think it would be very good. Should we do it in this PR or in a follow-up Issue?

Definitely not in a follow up issue since it was my initial request in https://github.com/poanetwork/tokenbridge/issues/167#issuecomment-515797106.

deployment/roles/common/tasks/repo.yml
```diff
@@ -1,25 +1,23 @@
...
-    git ls-tree -r HEAD --name-only
+    git ls-tree -r HEAD --name-only | sed '/^contracts$/d'
     cd contracts; git ls-tree -r HEAD --name-only | sed -e 's/^/contracts\//'
-  register: rsync_files
+  register: registered_files
   delegate_to: 127.0.0.1
   become: false
   args:
-    chdir: ".."
+    chdir: "{{ lookup('env', 'PWD') }}/.."
```
rzadp: This does not change the behaviour of the playbook (since `pwd/..` is the same as `..`), but for some reason Molecule messes up paths so that this is required to make this tasks work in Molecule tests.

deployment/roles/common/tasks/repo.yml
```diff
@@ -1,25 +1,23 @@
...
 - name: Copy the files
-  synchronize:
-    dest: "{{ bridge_path }}"
-    src: ../../../..
-    rsync_opts:
-      - "--include=\"{{ item }}\""
-      - "--exclude='*'"
-      - "-avz"
-  with_items: rsync_files.stdout_lines
+  copy:
```
rzadp: This does not create all the parent directories, that's why the previous task to create structure of directories is necessary.

rzadp:
### Current repo task

I had some problems with current [repo task](https://github.com/poanetwork/tokenbridge/blob/master/deployment/roles/common/tasks/repo.yml), which uses `synchronize` module.

The include/exclude patterns were not working correctly in the Molecule tests (although they worked fine when manually testing with Digital Ocean). Most likely it is same issue as [this](https://github.com/ansible/ansible/issues/23575).

### Second approach

I couldn't make the `synchronize` module to work correctly, so I tried the `copy` module. It turned out to be too slow, copying the files one-by-one took definitely too much time.

### Third approach

I archive the files using `tar`, and un-archive them on remote machine.

@akolotov @ArseniiPetrovich @phahulin Please review, especially these changes to the repo task.

akolotov:
It is related to 9df6ac0: @rzadp at which particular moment (on which package) `apt-get` exceed the timeout? We introduced installation of `rsync` package under #195 but it is not used now. If this particular change will be rolled back will it solve the issue with lengthy `apt-get`?

rzadp:
> It is related to [9df6ac0](https://github.com/poanetwork/tokenbridge/commit/9df6ac0703a5d497d4348df73639c9847dfd794f): @rzadp at which particular moment (on which package) `apt-get` exceed the timeout? We introduced installation of `rsync` package under #195 but it is not used now. If this particular change will be rolled back will it solve the issue with lengthy `apt-get`?

I don't know which package is causing this (the output does not show which package takes long time). I suspect `python`, because it has happened before, and I looked into it and turned out that `python` was downloaded at a very low speed. `rsync` is < 1MB so it can't be the problem

rzadp:
The problem went away after one or two days, when it happened before.

rzadp:
> the idea with copying the files seems to become complex. I approve the changes so far to merge them but we need to try another way. Is it possible to extract the current remote git repo url and the branch and pass them to the deployment scripts on the remote side?

Created https://github.com/poanetwork/tokenbridge/issues/200 for this.

Issue #202, opened 2019-09-06
remove dependency on `rsync`


akolotov:
As part of #195 usage of `rsync` was introduced to copy files from a local workcopy of the repo to a remote system during deployment. Later, `rsync` was replaced by copying archive with the local workcopy but the dependency to the `rsync` package still exists:
  * [deployment-e2e/Dockerfile](https://github.com/poanetwork/tokenbridge/blob/677bc50519aa04b81a237f01d2937e50e426f55d/deployment-e2e/Dockerfile)
  * [deployment/roles/common/tasks/dependencies.yml](https://github.com/poanetwork/tokenbridge/blob/01186d6aa86aeee5577bc3509e0eb4a78b752cc1/deployment/roles/common/tasks/dependencies.yml)

Issue #200, opened 2019-09-06
Investigate repository deployment method


rzadp:
https://github.com/poanetwork/tokenbridge/pull/194#issuecomment-526596743
https://github.com/poanetwork/tokenbridge/pull/194#pullrequestreview-284739828

This issue is to investigate the method for repository deployment.

1. Approach to `git clone` is not sufficient because it does not work with forked PRs
2. Approach to copy files, ignoring files not tracked by git, is complex
3. (to investigate) try approach to `git clone`, but supplying a repository address, so that it works with forked PRs

rzadp:
The result of the investigation is summarized here: https://github.com/poanetwork/tokenbridge/pull/205#issuecomment-531822019

PR #195, opened 2019-08-27
New way of deploying repository

rzadp:
- Closes #167 by changing the task cloning the repository, to just copy the files
- Removed now-obsolete `bridge_repo` and `bridge_repo_branch` configs
- Note that this PR is a forked PR from `rzadp/tokenbridge`, demonstrating working tests
- Related PR: #194 

- [x] TODO: Manually test deployment on DigitalOcean

PR #205, opened 2019-09-11
Old, modified way of deploying repository

rzadp:
- This PR reverts the changes done in #195 , with modification to make it work for forked PRs.
- Closes https://github.com/poanetwork/tokenbridge/issues/200
- Closes https://github.com/poanetwork/tokenbridge/issues/202
- Note that this is a forked PR


deployment-e2e/molecule/repo/converge.yml
```diff
@@ -3,6 +3,9 @@
   hosts: all
   become: true
   tasks:
+    - import_role:
+        name: ../../../deployment/roles/common
+        tasks_from: dependencies
```
rzadp: Need to add dependencies in this test scenario because otherwise there is no `git` installed on host machine

deployment/roles/common/tasks/repo.yml
```diff
@@ -1,34 +1,21 @@
...
-    - monorepo.tar.gz
-    - contracts.tar.gz
-
-- name: Untar the archives
+- name: Checkout proper branch
   shell: |
-    tar zxfv monorepo.tar.gz && rm monorepo.tar.gz
-    mkdir -p contracts && tar zxfv contracts.tar.gz -C ./contracts && rm contracts.tar.gz
+    # If this is a forked PR, the branch needs to be fetched in a special way
+    [ ! "CIRCLE_PR_NUMBER" = "" ] && git fetch --force origin -- "{{ bridge_repo_branch }}/head:remotes/origin/{{ bridge_repo_branch }}"
```
rzadp: `CIRCLE_PR_NUMBER` is only available on forked PR. So if it is available, we fetch the forked branch in a special way.

deployment-e2e/molecule.sh
```diff
@@ -2,8 +2,10 @@
 cd $(dirname $0)
 set -e # exit when any command fails
 
+CODEBASE_BRANCH=${CIRCLE_BRANCH-$(git symbolic-ref --short HEAD)}
```
akolotov: will it still be possible to run `e2e` deployment tests locally?

deployment/roles/common/tasks/repo.yml
akolotov: does it make sense to limit the number of downloaded artifacts by `--depth` parameter?

deployment-e2e/molecule.sh
rzadp: Good question.
- In a normal repository - yes
- In a forked repository - I will work on it now and make sure that it is possible

deployment/roles/common/tasks/repo.yml
rzadp: Good idea. Done in https://github.com/poanetwork/tokenbridge/pull/205/commits/a4caf18dcc33c33f14f92fc762e3488588159091

rzadp:
@akolotov I have trouble making it work on forked PRs. Turns out, the tests were passing before for the wrong reason. I found that I had two branches named the same way on both forked and original repository, and it was actually checking out the wrong one!

My current opinion is that approach to `git clone` is arguably more complex than the approach we have currently with tar archives.

When running `deployment-e2e` tests, and also when manually testing deployment on DigitalOcean/AWS, the approach to `git clone` needs to take into account 4 cases:

1. Locally on original repository
    - The local branch needs to be pushed to remote (in order to be `git cloned` later)
    - Need to make sure that in fact given branch is checked out and tested, and not the default `master`
2. CI testing PR from original repository
    - Need to make sure that in fact given branch is checked out and tested
3. Locally on forked repository
   - The local branch needs to be pushed to remote
    - Need to make sure that in fact given branch is checked out and tested
    - Need to make sure that in fact the forked repository is tested, and not the default `poanetwork/tokenbridge`
4. CI testing PR from forked repository
    - Need to make sure that in fact given branch from forked repository is checked out and tested
    - This branch must be fetched and checked-out in a special way

If something gets wrong, it is easy to overlook, because the codebase from original repository's `master` has high chance to just work. Just like I overlooked it with two branches.


I think that our current approach is better and simpler. The tasks to tar/untar are complex, but the principles are simple:

1. Locally, the code you see is deployed / tested. No matter if you have original repository, forked repository, or if you downloaded repository `.zip`
2. When testing original/forked PR, CI checks out proper code. This code gets tested.

In conclusion, I argue that it would be better to leave our current approach with copying repository as it is.
@akolotov What do you think?

akolotov:
> I think that our current approach is better and simpler. The tasks to tar/untar are complex, but the principles are simple:
> 
> 1. Locally, the code you see is deployed / tested. No matter if you have original repository, forked repository, or if you downloaded repository `.zip`
> 2. When testing original/forked PR, CI checks out proper code. This code gets tested.
> 
> In conclusion, I argue that it would be better to leave our current approach with copying repository as it is.
> @akolotov What do you think?

Thanks @rzadp for detailed explanation. Please add this comment to the issue that is opened to investigate possibility to switch back to `git clone`.

akolotov:
If nothing is going to be changed as part of this PR, please close it.

PR #198, opened 2019-09-05
Consistent variable naming

rzadp:
- Closes https://github.com/poanetwork/tokenbridge/issues/127
- `ERC20_TOKEN_ADDRESS ` will be removed in separate issue/PR: https://github.com/poanetwork/tokenbridge/issues/201

PR #207, opened 2019-09-11
[PR-into-PR] Reading environment variables in e2e-commons

rzadp:
- Addresses https://github.com/poanetwork/tokenbridge/pull/198#discussion_r323173768
- The variables from `e2e-commons/docker-compose.yml` are moved to `e2e-commons/components-envs/`
  - The `REACT_APP_` prefix has been removed, and is added by `ui` itself so it is consistent with deploying and running locally `ui`
- The existing `e2e-commons/envs` folder is renamed to `e2e-commons/contracts-envs` in order to make it clear which envs are used in oracle/ui/monitor, and which are used when deploying the contracts


ui/Dockerfile
```diff
@@ -16,6 +16,9 @@ RUN mv ./contracts/build ./ && rm -rf ./contracts/* ./contracts/.[!.]* && mv ./b
 COPY ./commons ./commons
 
 COPY ./ui ./ui
+ARG DOT_ENV_PATH=./ui/.env
```
akolotov: how did you check that it works?

rzadp: It’s tested by jobs in two ways:
1. The ultimate tests prove that it works with default path „ui/.env”.
The behaviour when deploying components is unchanged, because „ui/.env” is copied anyway in previous step.
2. Passing e2e tests prove that it works with overriden path to .env

akolotov: Thanks!

Issue #201, opened 2019-09-06
Remove ERC20_TOKEN_ADDRESS from env variables


rzadp:
https://github.com/poanetwork/tokenbridge/issues/127#issuecomment-526509190

> Please note that ERC20_TOKEN_ADDRESS is not need to be specified any more since this information could be extracted from the blockchain.

Separate issue to do it in a separate PR.

PR #211, opened 2019-09-13
Remove ERC20_TOKEN_ADDRESS from env variables

rzadp:
- Closes https://github.com/poanetwork/tokenbridge/issues/201
- Also removing `BRIDGEABLE_TOKEN_ADDRESS ` which does not seem to be used anywhere
- `ERC20_TOKEN_ADDRESS` value is determined in `initialChecks` in oracle, which required me to change e2e setup to perform `deploy` first, and after that spin up `oracle`


oracle/scripts/initialChecks.js
```diff
@@ -1,15 +1,33 @@
...
-    const foreignWeb3 = new Web3(new Web3.providers.HttpProvider(COMMON_FOREIGN_RPC_URL))
-    const bridgeTokenContract = new foreignWeb3.eth.Contract(ERC677_BRIDGE_TOKEN_ABI, ERC20_TOKEN_ADDRESS)
+    const bridge = new foreignWeb3.eth.Contract(FOREIGN_ERC_TO_ERC_ABI, COMMON_FOREIGN_BRIDGE_ADDRESS)
+    result.ERC20_TOKEN_ADDRESS = await bridge.methods.erc20token().call()
+  } else if (ORACLE_BRIDGE_MODE === 'ERC_TO_NATIVE') {
+    const bridge = new foreignWeb3.eth.Contract(FOREIGN_ERC_TO_NATIVE_ABI, COMMON_FOREIGN_BRIDGE_ADDRESS)
+    result.ERC20_TOKEN_ADDRESS = await bridge.methods.erc20token().call()
+  } else if (ORACLE_BRIDGE_MODE === 'NATIVE_TO_ERC') {
+    const bridge = new foreignWeb3.eth.Contract(FOREIGN_NATIVE_TO_ERC_ABI, COMMON_FOREIGN_BRIDGE_ADDRESS)
+    result.ERC20_TOKEN_ADDRESS = await bridge.methods.erc677token().call()
```
akolotov: From my perspective `ERC20_TOKEN_ADDRESS` should not be used in the `native-to-erc` mode.

oracle/config/affirmation-request-watcher.config.js
```diff
@@ -26,7 +26,7 @@ module.exports =
         ...baseConfig.bridgeConfig,
         ...baseConfig.foreignConfig,
         event: 'Transfer',
-        eventContractAddress: process.env.ERC20_TOKEN_ADDRESS,
+        eventContractAddress: initialChecks.ERC20_TOKEN_ADDRESS,
```
akolotov: let's keep consistency in the variables names: use the 'camel'-style (as `initialChecks.foreignERC` above) instead of `_`.

Also please consider to rename the variable to `bridgeableTokenAddress`.

rzadp:
I just realized there might by a chicken-and-egg problem with this.

1. For `ERC_TO_ERC`, we call `getTokenType` to the bridgeableToken to determine whether it is `ERC677` or `ERC20`
2. In order to call `getTokenType` to the bridgeableToken, we need it's address
3. We want to get it's address using `bridge.methods.erc20token()` or `bridge.methods.erc677token()`, instead of environment variable `ERC20_TOKEN_ADDRESS `
4. We don't know which method do use because we don't know whether it is `ERC677` or `ERC20`

I'm sure we had this conversation before but I don't remember what was the conclusion.
@akolotov @patitonar Can you provide some guidance?

Another problem is that the tests are passing for this PR, although now I see that it would probably fail for `ERC_TO_ERC` with `ERC20_EXTENDED_BY_ERC677=true`. It is because it is not covered by ultimate tests, I never got around to do https://github.com/poanetwork/tokenbridge/issues/56. @akolotov Should I work on it?

rzadp:
@patitonar Got it, thanks!

oracle/scripts/initialChecks.js
rzadp: Updated, checking the token address only for `ERC_TO_ERC` and `ERC_TO_NATIVE`

oracle/config/affirmation-request-watcher.config.js
rzadp: Done, renamed `ERC20_TOKEN_ADDRESS` to `bridgeableTokenAddress`, in other places in code as well.

akolotov: thanks

akolotov:
> Another problem is that the tests are passing for this PR, although now I see that it would probably fail for ERC_TO_ERC with ERC20_EXTENDED_BY_ERC677=true. It is because it is not covered by ultimate tests, I never got around to do #56. @akolotov Should I work on #56 as well?

Thanks for heads up. The plan is to address #56 later, not now.

rzadp:
Having clarified it https://github.com/poanetwork/tokenbridge/pull/211#issuecomment-531246400 I now think that the current solution would work for `ERC_TO_ERC` with `ERC20_EXTENDED_BY_ERC677=true` as well.

akolotov:
@patitonar could you review the changes too?

PR #222, opened 2019-10-23
Fix suggested gas price in trasaction for ui production build

patitonar:
When using the production build of the UI it was found that a huge number was suggested as gas price for the transaction.
image

The problem is present only in the production build of the UI, the `toHex` method of `web3-utils` returns a wrong result only in the production build when the number provided is of the type `bn.js` which is used by web3. 

Example of result in dev build: `0x12a05f200`
Example of result in production build: `0x2231326130356632303022`

There is an issue created but it was not solved https://github.com/ethereum/web3.js/issues/2775. 

Passing the value as string works as workaround and avoids this problem.

### DIFF
--- a/ui-e2e/package.json
+++ b/ui-e2e/package.json
@@ -4,9 +4,9 @@
   "license": "MIT",
   "private": true,
   "devDependencies": {
-    "selenium-webdriver": "3.6.0",
+    "chromedriver": "^77.0.0",
     "mocha": "^5.2.0",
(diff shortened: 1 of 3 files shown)


patitonar:
Updated chrome driver version used in e2e tests that was failing.

The reason the issue with `toHex` wasn't detected by the e2e tests is because we are using the development build of the UI, I think we should create an issue to implement the changes to use the  production build of the UI app in order to detect situations like this.

PR #224, opened 2019-11-01
Support token migration

patitonar:
This PR includes changes explained here https://forum.poa.network/t/migration-to-mcd-for-xdai-bridge/3136

- The oracle `Transfer` watcher will discard transfers that includes a `TokensSwapped` event in the same transaction.

- The monitor was updated to also filter the `Transfer` events that includes a `TokensSwapped` event in the same transaction and to get `Transfer` events of previous token used to correctly calculate the results.

Tested the implementation with MakerDAO contracts in Sokol - Kovan testnets:

Sai Address: `0xc4375b7de8af5a38a93548eb8453a498222c4ff2`
Dai Address: `0x1d7e3a1a65a367db1d1d3f51a54ac01a2c4c92ff`
ScdMcdMigration Address: `0x97cb5a9abcdbe291d0cd85915fa5b08746fe948a`

HomeBridge: `0x6E4b2262CFD96Af188C2c3Aae21a24a53A88E9F4`
ForeignBridge: `0x4F7925718A2492BD41D77D9F81552535E72A2090`
migrateToMCD call: `0x332c540a8c711a680167088fdf3732399f33707f07a08f56dde41a4f0b9377ac`


monitor/utils/events.js
```diff
@@ -96,11 +97,40 @@ async function main(mode) {
...
+
     // Get transfer events that didn't have a UserRequestForAffirmation event in the same transaction
-    const directTransfers = transferEvents.filter(
+    let directTransfers = transferEvents.filter(
       e => foreignToHomeRequests.findIndex(t => t.referenceTx === e.referenceTx) === -1
     )
 
+    // filter transfer that is part of a token swap
+    directTransfers = directTransfers.filter(
+      e => tokensSwappedEvents.findIndex(t => t.referenceTx === e.referenceTx) === -1
```
akolotov: In this case `TokensSwapped` will affect an occurrence of `Transfer` even if `Transfer` is not a result of `mint` call. Consider to check the zero address in the field `from` of the corresponding `Transfer` at least.

monitor/utils/events.js
```diff
@@ -96,11 +97,40 @@ async function main(mode) {
...
+    const tokensSwappedEvents = (await getPastEvents(foreignBridgeSwapped, {
+      event: 'TokensSwapped',
+      fromBlock: MONITOR_FOREIGN_START_BLOCK,
+      toBlock: foreignBlockNumber
+    })).map(normalizeEvent)
+
+    // Get transfer events for each previous erc20
+    await Promise.all(
+      tokensSwappedEvents.map(async swap => {
+        const previousERC20 = new web3Foreign.eth.Contract(ERC20_ABI, swap.recipient)
```
akolotov: `recipient` is quite misleading. Does it make sense to implement another normalization method different from `normalizeEventInformation`?

monitor/utils/events.js
```diff
@@ -96,11 +97,40 @@ async function main(mode) {
...
+
+        const previousTransferEvents = (await getPastEvents(previousERC20, {
```
akolotov: Imagine that we have dozens of `TokensSwapped` events for the same pair of tokens. E.g. for the case when the bridge contracts will initiate the swap process every time when new SCD tokens appears on the balance. It will lead that for every instance of `TokensSwapped` the monitor will filter `Transfer` again and again for the same token contract address getting the same results. Consider to implement functionality to request `Transfer` events only for unique token contract addresses.

monitor/utils/events.js
```diff
@@ -96,11 +97,40 @@ async function main(mode) {
...
+          event: 'Transfer',
```
akolotov: What if some `Transfer` appears as result of a token swap. E.g. `<some first token>` -> `some another token` -> `some third token`? Seems that:
  - such `Transfer` should no be taken into account
  - a chain of swap operations should be investigated recursively

What do you think?

oracle/src/events/processTransfers/index.js
```diff
@@ -55,6 +57,15 @@ function processTransfersBuilder(config) {
           return
         }
 
+        const existsTokensSwappedEvent = receipt.logs.some(e => e.topics[0] === tokensSwappedHash)
+
+        if (from === ZERO_ADDRESS && existsTokensSwappedEvent) {
```
akolotov: Will it be more safe to also check that the emitter of `TokensSwapped` is the bridge contract address?

monitor/utils/events.js
```diff
@@ -96,11 +97,40 @@ async function main(mode) {
       }
     })).map(normalizeEvent)
 
+    const foreignBridgeSwapped = new web3Foreign.eth.Contract([TOKENS_SWAPPED_EVENT_ABI], COMMON_FOREIGN_BRIDGE_ADDRESS)
+    const tokensSwappedEvents = (await getPastEvents(foreignBridgeSwapped, {
+      event: 'TokensSwapped',
```
akolotov: Will it be more safe to also check that the emitter of `TokensSwapped` is the bridge contract address?

monitor/utils/events.js
patitonar: I'm not sure I fully understood the scenario.
In the code we get all the `TokensSwapped` events, then for each token address that was swapped we get all the transfers to the bridge here:
https://github.com/poanetwork/tokenbridge/blob/a52b23cc0f4aa7e58bda9cf034d9059cf2192c67/monitor/utils/events.js#L112-L124

Then, once we have all the transfers from previous and current tokens addresses, we filter them to exclude those that were emitted as a result of a swap:
https://github.com/poanetwork/tokenbridge/blob/a52b23cc0f4aa7e58bda9cf034d9059cf2192c67/monitor/utils/events.js#L131-L137

monitor/utils/events.js
patitonar: Yes, applied the change https://github.com/poanetwork/tokenbridge/pull/224/commits/a52b23cc0f4aa7e58bda9cf034d9059cf2192c67

oracle/src/events/processTransfers/index.js
patitonar: Yes, applied the change https://github.com/poanetwork/tokenbridge/pull/224/commits/d0117edf6bc555d5690b1b90d5596a9b89114f14

commons/abis.js
```diff
@@ -48,6 +48,29 @@ const ERC20_BYTES32_ABI = [
   }
 ]
 
+const TOKENS_SWAPPED_EVENT_ABI = {
```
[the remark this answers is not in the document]
patitonar: That's correct, updated the code to get it as part of `FOREIGN_ERC_TO_NATIVE_ABI`

monitor/utils/events.js
patitonar: Thanks, added the check for `from` parameter

monitor/utils/events.js
patitonar: You're right, I removed the normalization, using the event directly is cleaner https://github.com/poanetwork/tokenbridge/pull/224/commits/a52b23cc0f4aa7e58bda9cf034d9059cf2192c67

monitor/utils/events.js
patitonar: Updated the code to deduplicate token contract addresses in order to no process them twice https://github.com/poanetwork/tokenbridge/pull/224/commits/a52b23cc0f4aa7e58bda9cf034d9059cf2192c67

monitor/utils/events.js
akolotov: Yes, you are right. I have missed the fact that you get `TokensSwapped` events for `COMMON_FOREIGN_BRIDGE_ADDRESS`.

So, now I see that the following sequence of the events:
```
B.Transfer(0, bridge, <some value>
bridge.TokensSwapped(A, B, <some value>)
C.Transfer(0, bridge, <some value>
bridge.TokensSwapped(B, C, <some value>)
D.Transfer(0, bridge, <some value>
bridge.TokensSwapped(C, D, <some value>)
```
will be handled correctly.

commons/abis.js
akolotov: You have removed this ABI usage. Don't think it is needed now.

patitonar: thanks, I removed it https://github.com/poanetwork/tokenbridge/pull/224/commits/ec887c7c3ca8bd2f028a01af9ad5908d3f52cdd8

PR #244, opened 2019-11-29
Support two tokens deposit requests in Oracle 

### LINKED ISSUE #243: Support of two tokens (SCD and MCD) deposit requests by the oracle
This issue is to reflect in the oracle the corresponding changes made under https://github.com/poanetwork/tokenbridge-contracts/issues/318. 

Support of the additional ERC20 token on the oracle side requires a new watcher to listen `Transfer` events raised by this token contract.

It can be done by copying the functionality of the existing `processTransfer` watcher but instead of getting the token address by the `erc20token` method this watcher will work with the token address received from `halfduplexErc20token`.

The next important difference is that the watcher will try to execute the SCD deposit requests only if `isTokenSwapAllowed` described above returns `true`. This method expects the timestamp as a parameter so, the watcher will use the timestamp of [the block it uses in the filter to receive the events](https://github.com/poanetwork/tokenbridge/blob/303b02f3ca3dc230fa8b63b90c9235806a9ef508/oracle/src/watcher.js#L137). In fact, the number of this block is calculated as subtraction of the required block confirmations from the latest chain block.

Few approaches were considered how to `swapTokens` could be invoked after scheduling the job to `executeAffirmation`. As the result the following approach was chosen:
  * the watcher of `Transfer` events from the SCD token contract creates a new job in the Rabbit message queue to identify if `swapTokens` should be called before scheduling invocation of `executeAffirmation`. [The block number that was used in the filter](https://github.com/poanetwork/tokenbridge/blob/303b02f3ca3dc230fa8b63b90c9235806a9ef508/oracle/src/watcher.js#L137) to receive `Transfer` events as the end of the block range is associated with the new job.
  * a new worker receives the job and the associated block number from the Rabbit message queue.
  * it checks if the current SCD balance for the bridge contract is above the minimal balance by calling `isHDTokenBalanceAboveMinBalance`
  * another check is needed to understand if the validator running the node with the oracle is on duty to execute tokens swap. The check requires the call of `isValidatorDuty`.
  * since it could be several `Transfer` events caught by the watcher for a new block, only one event should trigger `swapTokens`. That's why the worker compares the block number received with the job and the block number in the Redis database stored by the previous iteration of the worker. If the block numbers are equal, the job is skipped.
  * now the method `swapTokens` could be estimated and scheduled through the Rabbit message queue for the worker that is responsible for sending transactions to the Ethereum Mainnet (Foreign chain). If some errors occur during the call estimation, the worker should report about the error similar to current reports from existing watchers of events
  * if the invocation of `swapTokens` is scheduled correctly the block number associated with the job stored in the Redis database.

Since the new two workers are going to be added to the oracle node the deployment scripts should be updated to:
  - run the new workers automatically as part of the bridge service
  - forward logs produced by the new workers to a remote syslog server

patitonar:
Closes #243 

This PR includes:
- [x] Transfer watcher for `halfduplexErc20token`
- [x] Add support for a new type of workers with the swap tokens worker implementation
- [x] Update deployment playbooks to include the new workers in docker deployment and logs configuration
- [x] Update parity spec on foreign side to include deployed mock contracts in specific addresses to use in e2e tests (sai, dai, daiAdapter, saiTop, migration contracts) 
- [x] Add e2e tests for new functionality in oracle


deployment/roles/oracle/tasks/logging.yml
```diff
@@ -1,41 +1,9 @@
...
-- name: Set logger to remote server for extended file
-  set_fact:
-    docker_compose_extended_parsed: "{{ docker_compose_extended_parsed |combine({'services': {item: {'logging': {'driver': 'syslog','options': {'tag': '{{.Name}}/{{.ID}}'}}}}}, recursive=True) }}"
-  with_items: "{{ docker_compose_extended_parsed.services }}"
-
-- name: Write new docker-compose-extended file
-  copy:
-    content: "{{ docker_compose_extended_parsed | to_yaml }}"
-    dest: "{{ bridge_path }}/oracle/docker-compose-transfer.yml"
+- include_tasks: logging_remote.yml
```
akolotov: in fact, the actions performing by this task are configuring the docker-compose facility to forward logs to the syslog service on the host machine. Do you think it make sense to rename the file in `logging_by_syslog.yml`?

oracle/src/workers/swapTokens.js
```diff
@@ -0,0 +1,92 @@
...
+      })
+
+      logger.debug({ gasEstimate }, 'Gas estimated')
+    } catch (e) {
+      if (e instanceof HttpListProviderError) {
+        const errorMsg = 'RPC Connection Error: swapTokens Gas Estimate cannot be obtained.'
+        logger.error(e, errorMsg)
+        throw new Error(errorMsg)
+      } else {
+        logger.error(e, 'Unknown error while processing transaction')
```
akolotov: Actually, based on the `swapTokens()` method functionality, there is at least one possible case that can be handled and reported:
- the HD token balance is zero. Could happen if a new block appears between `isHDTokenBalanceAboveMinBalance()` call and `estimateGas` invocation.

Other cases relate to internal functionality of the token and the migration contracts so we cannot identify the reason of revert in these cases for sure.

oracle/src/workers/swapTokens.js
```diff
@@ -0,0 +1,92 @@
...
+
+    logger.debug(`Checking if half duplex balance is above the threshold`)
+    const hdTokenBalanceAboveMinBalance = await foreignBridge.methods.isHDTokenBalanceAboveMinBalance().call()
+
+    if (!hdTokenBalanceAboveMinBalance) {
+      logger.info(`Token swap discarded because half duplex balance is below the threshold`)
+      return txToSend
+    }
+
+    logger.debug(`Getting block`)
```
akolotov: Do you think it makes sense to report the block timestamp in the log?

oracle/src/events/processHalfDuplexTransfers/index.js
```diff
@@ -0,0 +1,132 @@
...
+      .map(transfer => async () => {
+        const { from, value } = transfer.returnValues
+
+        const logger = rootLogger.child({
+          eventTransactionHash: transfer.transactionHash
+        })
+
+        logger.info({ from, value }, `Processing transfer ${transfer.transactionHash}`)
+
+        const block = await web3Foreign.eth.getBlock(blockNumber)
```
akolotov: Do you think it makes sense to report the block timestamp in the log?

oracle/config/half-duplex-transfer-watcher.config.js
```diff
@@ -0,0 +1,38 @@
...
+if (!transferWatcherRequired) {
+  console.error(`Transfer watcher not required for bridge mode ${process.env.ORACLE_BRIDGE_MODE}`)
+  process.exit(EXIT_CODES.WATCHER_NOT_REQUIRED)
+}
+
+module.exports = {
+  ...baseConfig.bridgeConfig,
+  ...baseConfig.foreignConfig,
+  event: 'Transfer',
+  eventContractAddress: initialChecks.halfDuplexTokenAddress,
```
akolotov: we need to consider the case when `halfDuplexTokenAddress` is not defined. There are two possible scenarios:
- the oracle has been updated before the upgrade of the bridge contracts happens
- the oracle is used with a token different from Dai

oracle/src/events/processHalfDuplexTransfers/index.js
```diff
@@ -0,0 +1,132 @@
...
+          : false
+
+        if (from === ZERO_ADDRESS && existsTokensSwappedEvent) {
+          logger.info(
+            `Transfer event discarded because token swap is detected in transaction ${transfer.transactionHash}`
+          )
+          return
+        }
+
+        let gasEstimate
```
akolotov: we need to consider a scenario when the oracle has been upgraded before `migrateToMCD` is invoked. In this case both `erc20token()` and `halfDuplexErc20token()` will return the same. So, each oracle will try to confirm the deposit requests twice.

oracle/src/workers/swapTokens.js
```diff
@@ -0,0 +1,92 @@
...
+
+    logger.debug(`Checking if is validator duty`)
+    const validatorDuty = await validatorContract.methods.isValidatorDuty(config.validatorAddress).call()
+
+    if (!validatorDuty) {
+      logger.info(`Token swap discarded because is not validator duty`)
+      return txToSend
+    }
+
+    logger.debug(`Checking if half duplex balance is above the threshold`)
```
akolotov:
```suggestion
    logger.debug(`Checking if half duplex token balance is above the threshold`)
```

oracle/config/half-duplex-transfer-watcher.config.js
patitonar: I'm trying to look for approaches for this. At the moment of starting the halfDuplexToken watcher we could find the following scenarios:
- `halfDuplexErc20token` method does not exist (contract were not upgraded yet or is not dai).
- `halfDuplexErc20token` method exists but it returns the same address as the `erc20token` method (contracts were updated but `migrateToMCD` wasn't invoked).

In both scenarios, halfDuplexToken watcher and the swap tokens worker should not perform any actions. The swap tokens worker is subscribed to a queue, so it will be idle while no watcher pushes a job to that queue.

Regarding the halfDuplexToken watcher, we can end the process after checking the existence of the methods but docker will try to restart them every time and keep the watcher in a loop of starting the app, performing checks, ending the app.

We can use a different docker-compose file if the conditions are not met to avoid including the new watcher and worker, but after performing the contract upgrade and migration, it will require more than just restarting the oracle component.  

One idea could be to add a functionality that a worker could be set to sleep forever if some conditions are met.

For example, on the initial checks when the watcher is starting, if one of the two scenarios are detected, we can set an attribute and then set the watcher to sleep instead of listening to the events.

So after the contracts are upgraded and `migrateToMCD` is invoked, restarting the components will make the watcher start working correctly.

deployment/roles/oracle/tasks/logging.yml
patitonar: Sure, renamed https://github.com/poanetwork/tokenbridge/pull/244/commits/a4eb446f7bf7571725261a277e6e3fdf6529a98e

oracle/src/workers/swapTokens.js
patitonar: Yes, added timestamp log https://github.com/poanetwork/tokenbridge/pull/244/commits/ab2c0ea120fa1efe6af213ce9400b7cb0a291bee

oracle/src/events/processHalfDuplexTransfers/index.js
patitonar: Yes, added timestamp log https://github.com/poanetwork/tokenbridge/pull/244/commits/ab2c0ea120fa1efe6af213ce9400b7cb0a291bee

oracle/src/workers/swapTokens.js
patitonar: Good catch! Added special check for zero balance https://github.com/poanetwork/tokenbridge/pull/244/commits/5583ea8b6bba49a0812247f5e75ba0edf2c8b7ca

oracle/src/workers/swapTokens.js
```diff
@@ -0,0 +1,111 @@
...
+          halfDuplexTokenContract = new web3Foreign.eth.Contract(ERC20_ABI, halfDuplexErc20Token)
+        }
+
+        const balance = web3Foreign.utils.toBN(
+          await halfDuplexTokenContract.methods.balanceOf(config.foreignBridgeAddress).call()
+        )
+        logger.debug({ balance: balance.toString() }, 'Half duplex token bridge balance obtained')
+
+        if (balance.isZero()) {
+          logger.info(`Gas estimate failed because half duplex balance is zero. Token swap is discarded.`)
```
akolotov:
```suggestion
          logger.info(`Gas estimate failed because half duplex token balance is zero. Tokens swap is discarded.`)
```

oracle/config/half-duplex-transfer-watcher.config.js
akolotov: thanks for analysis! Did you consider a way to modify [the `main` function](https://github.com/poanetwork/tokenbridge/blob/ab2c0ea120fa1efe6af213ce9400b7cb0a291bee/oracle/src/watcher.js#L131) as so it check if the watcher is for the half duplex token transfers and if the both `halfDuplexErc20token()` and `erc20token()` return the same address (or `halfDuplexErc20token()` fails) it does `return`?

patitonar: Thanks for the suggestion! Please check https://github.com/poanetwork/tokenbridge/pull/244/commits/8fa715089b5a4e6e142b28c46990cf940f91cc27

akolotov: Awesome! Does it mean that we need to restart the oracle after the `migrateToMCD` call?

patitonar: That's correct. After calling `migrateToMCD` we need to restart the oracle. After the restart, the `Transfer-watcher` will start listening to events of the updated `erc20token()` address and the `halfDuplexToken-transfer-watcher` will start to listen to events of the `halfDuplexErc20token()` address

akolotov: It is always a pain to synchronize validators to perform some actions. That's why do you think it is a bad idea to move

https://github.com/poanetwork/tokenbridge/blob/8fa715089b5a4e6e142b28c46990cf940f91cc27/oracle/scripts/initialChecks.js#L20-L30

to [the main function](https://github.com/poanetwork/tokenbridge/blob/8fa715089b5a4e6e142b28c46990cf940f91cc27/oracle/src/watcher.js#L133)?

patitonar: This will give the oracle some flexibility which will be good to avoid restarting the services. The only bad side is that it will add some overhead (RPC requests to get addresses) for the specific watchers. I'll update the code to support running checks before getting the events and update the variables needed

patitonar: Updated the implementation and also added an e2e test case to cover that the oracle works before and after the migration. Please check https://github.com/poanetwork/tokenbridge/pull/244/commits/3cd53f7bdab5a4f203ee0d8a94ba848c5e9a0f78

oracle/src/events/processTransfers/tokenState.js
```diff
@@ -0,0 +1,20 @@
+async function getTokensState(bridgeContract) {
```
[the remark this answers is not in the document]
patitonar: Yes, it does make sense to move it to `utils`, thanks. https://github.com/poanetwork/tokenbridge/pull/244/commits/0829c95561be85280aa55b5a87308f9ad562c8bd

PR #245, opened 2019-12-05
Support two tokens deposits in monitor

patitonar:


monitor/getBalances.js
```diff
@@ -83,6 +83,28 @@ async function main(bridgeMode) {
     const foreignBridge = new web3Foreign.eth.Contract(FOREIGN_ERC_TO_NATIVE_ABI, COMMON_FOREIGN_BRIDGE_ADDRESS)
     const erc20Address = await foreignBridge.methods.erc20token().call()
     const erc20Contract = new web3Foreign.eth.Contract(ERC20_ABI, erc20Address)
+    let foreignHalfDuplexErc20Balance = 0
+    let displayHalfDuplexToken = false
```
akolotov: Do you think it makes sense to implement an opposite logic: first check if the bridge supports half duplex token and only after that check if the token swap allowed?

monitor/getBalances.js
```diff
@@ -91,24 +113,38 @@ async function main(bridgeMode) {
...
-    const mintedCoins = await blockRewardContract.methods.mintedTotally().call()
+    const mintedCoins = await blockRewardContract.methods.mintedTotallyByBridge(COMMON_HOME_BRIDGE_ADDRESS).call()
     logger.debug('calling homeBridge.methods.totalBurntCoins')
     const burntCoins = await homeBridge.methods.totalBurntCoins().call()
 
     const mintedCoinsBN = new BN(mintedCoins)
     const burntCoinsBN = new BN(burntCoins)
     const totalSupplyBN = mintedCoinsBN.minus(burntCoinsBN)
     const foreignErc20BalanceBN = new BN(foreignErc20Balance)
+    const halfDuplexErc20BalanceBN = new BN(foreignHalfDuplexErc20Balance)
```
akolotov: will it behave properly if the bridge does not support half duplex token?

monitor/getBalances.js
```diff
@@ -91,24 +113,38 @@ async function main(bridgeMode) {
...
+      .toFixed()
+
+    let foreign = {
+      erc20Balance: Web3Utils.fromWei(foreignErc20Balance)
+    }
+
+    if (displayHalfDuplexToken) {
+      foreign = {
+        ...foreign,
+        halfDuplexErc20Balance: Web3Utils.fromWei(foreignHalfDuplexErc20Balance)
```
akolotov: how about an idea to display the half duplex token balance after the emergency shutdown but in another field like `halfDuplexErc20BalanceAfterES`? This value will not be taken into account for the `diff` calculation but will be reported in the monitor's artefacts.

monitor/getBalances.js
patitonar: Yes, I think it's better in the way you described. Updated the logic https://github.com/poanetwork/tokenbridge/pull/245/commits/22b1bf162377f7033d40ada2ba4a7f2866a583a8

monitor/getBalances.js
patitonar: Yes, it will behave properly since the variable is initialized in `zero` if half duplex token is not supported yet so it won't have an impact in the diff. In the updated implementation it is more clear

monitor/getBalances.js
patitonar: That sounds good to me, I added the changes https://github.com/poanetwork/tokenbridge/pull/245/commits/22b1bf162377f7033d40ada2ba4a7f2866a583a8

monitor/getBalances.js
akolotov: thanks!

PR #253, opened 2019-12-26
Support several monitors on the same system

### LINKED ISSUE #138: Host of several monitors on the same system
During the review of the PR [Monitor deployment playbook](https://github.com/poanetwork/tokenbridge/pull/133) the following suggestion appears:

```
As a suggestion for discussion, would it be difficult to have this playbook support
multiple monitorings for multiple networks running on the same host? It would
require to use different ports, give different names to docker containers, services
and have multiple folders with separate .env.
This is how we currently run it (monitoring for 3 bridges on one host). 
```

So, this issue is to perform research and implement the best way to monitor operations on several bridges on the same host.

Possible approaches:
1. Separate docker container for a separate bridge. Requires monitors to be on separate ports and run containers with different names.
2. Separate end points for separate bridges. Requires changes on the monitor configuration level and the monitor workflow.
3. ???

patitonar:
Closes #138 

Introduces a new parameter `BRIDGE_NAME` for the deployment playbooks of the monitor. It is used to create different names for:

- The folder where the repo and `.env` is stored
- The service created to run de monitor
- The docker containers
- The cron job
- Some config files that are stored in the system


patitonar:
The approach looks good to me, I think you covered all the required changes. 

> The deployment scripts will check if the monitor has been already deployed on the system.

Do you have any suggestions on how to detect if the monitor has been already deployed? One option could be to check if the service `tokenbridge-monitor` is in place.

akolotov:
> Do you have any suggestions on how to detect if the monitor has been already deployed? One option could be to check if the service `tokenbridge-monitor` is in place.

Do you think it makes sense to better check existence of `docker-compose.yml` in the corresponding directory?

patitonar:
> Do you think it makes sense to check existence of `docker-compose.yml` in the corresponding directory?

Yes, I think that is a good option.

patitonar:
Closed in favor of #262

PR #254, opened 2019-12-27
Fix home gas price fallback value usage in monitor

### LINKED ISSUE #237: Disable validator balances checks in monitor for xDai-like networks
It should be possible (via configuration option or based on bridge type) to disable this check
https://github.com/poanetwork/tokenbridge/blob/84ecfc30d91dc9d94c530a7af3c965a0de83a65f/monitor/index.js#L55-L60
for networks like xDai where validator balances are always `0` on the home side.

patitonar:
Closes #237

If `COMMON_HOME_GAS_PRICE_FALLBACK` is set to `0` the `leftTx` attribute will be set to a very big number as specified in this change https://github.com/poanetwork/tokenbridge/pull/142#discussion_r302644176 , so then `results.homeOk` will be true when validators have `0` balance as in xDai Chain.

Most of the logic was in place, this PR only fixes the precedence order of the gas price parameters used in the calculation

### DIFF
--- a/monitor/validators.js
+++ b/monitor/validators.js
@@ -87,9 +87,10 @@ async function main(bridgeMode) {
   const homeVBalances = {}
 
   logger.debug('calling home getGasPrices')
-  const homeGasPrice =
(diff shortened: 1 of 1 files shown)


patitonar:
Thanks for the explanation! Closed in favor of https://github.com/poanetwork/tokenbridge/pull/259

PR #259, opened 2019-12-30
Add support to disable validator balances check in monitor

### LINKED ISSUE #237: Disable validator balances checks in monitor for xDai-like networks

patitonar:
Closes #237

If the variables `MONITOR_VALIDATOR_HOME_TX_LIMIT` or `MONITOR_VALIDATOR_FOREIGN_TX_LIMIT` are set to `0` or undefined, the check for validators balances is disabled.

### DIFF
--- a/monitor/index.js
+++ b/monitor/index.js
@@ -5,6 +5,8 @@ const { isV1Bridge } = require('./utils/serverUtils')
 
 const app = express()
 
+const MONITOR_VALIDATOR_HOME_TX_LIMIT = Number(process.env.MONITOR_VALIDATOR_HOME_TX_LIMIT) || 0
+const MONITOR_VALIDATOR_FOREIGN_TX_LIMIT = Number(process.env.MONITOR_VALIDATOR_FOREIGN_TX_LIMIT) || 0
 const MONITOR_TX_NUMBER_THRESHOLD = Number(process.env.MONITOR_TX_NUMBER_THRESHOLD) || 100
 console.log('MONITOR_TX_NUMBER_THRESHOLD = ' + MONITOR_TX_NUMBER_THRESHOLD)
 
@@ -52,18 +54,25 @@ app.get('/validators', async (req, res, next) => {
     const results = await readFile('./responses/validators.json')
     results.homeOk = true
     results.foreignOk = true
-    for (const hv in results.home.validators) {
-      if (results.home.validators[hv].leftTx < MONITOR_TX_NUMBER_THRESHOLD) {
-        results.homeOk = false
-        break
+
+    if (MONITOR_VALIDATOR_HOME_TX_LIMIT) {
+      for (const hv in results.home.validators) {
+        if (results.home.validators[hv].leftTx < MONITOR_TX_NUMBER_THRESHOLD) {
+          results.homeOk = false
+          break
+        }
       }
     }
-    for (const hv in results.foreign.validators) {
-      if (results.foreign.validators[hv].leftTx < MONITOR_TX_NUMBER_THRESHOLD) {
-        results.foreignOk = false
-        break
+
+    if (MONITOR_VALIDATOR_FOREIGN_TX_LIMIT) {
+      for (const hv in results.foreign.validators) {
+        if (results.foreign.validators[hv].leftTx < MONITOR_TX_NUMBER_THRESHOLD) {
+          results.foreignOk = false
+          break
+        }
       }
     }
+
     results.ok = results.homeOk && results.foreignOk
     res.json(results)
   } catch (e) {
--- a/monitor/validators.js
+++ b/monitor/validators.js
@@ -136,11 +136,9 @@ async function main(bridgeMode) {
   await asyncForEach(homeValidators, async v => {
     const balance = await web3Home.eth.getBalance(v)
     if (MONITOR_VALIDATOR_HOME_TX_LIMIT) {
-      const leftTx = homeTxCost.isZero()
-        ? 999999
-        : Web3Utils.toBN(balance)
-            .div(homeTxCost)
-            .toString(10)
+      const leftTx = Web3Utils.toBN(balance)
+        .div(homeTxCost)
+        .toString(10)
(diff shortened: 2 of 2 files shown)


monitor/index.js
```diff
@@ -5,6 +5,8 @@ const { isV1Bridge } = require('./utils/serverUtils')
 
 const app = express()
 
+const MONITOR_VALIDATOR_HOME_TX_LIMIT = Number(process.env.MONITOR_VALIDATOR_HOME_TX_LIMIT) || 0
```
akolotov: I think we should not request the gas price and fill `*VBalances` in `validators.js` if the corresponding `MONITOR_VALIDATOR_*_TX_LIMIT` is configured as zero. We can reduce network traffic if these requests are not sent.

monitor/validators.js
```diff
@@ -65,76 +65,99 @@ async function main(bridgeMode) {
...
     from: MONITOR_HOME_START_BLOCK,
     to: homeBlockNumber,
     logger
-  })
+  })).map(web3Home.utils.toChecksumAddress)
 
-  const homeBalances = {}
-  logger.debug('calling asyncForEach homeValidators homeBalances')
-  await asyncForEach(homeValidators, async v => {
-    homeBalances[v] = Web3Utils.fromWei(await web3Home.eth.getBalance(v))
```
patitonar: Removed this calls to get the validator balance on home since they were not being used

monitor/validators.js
```diff
@@ -65,76 +65,99 @@ async function main(bridgeMode) {
   const foreignBridgeValidators = new web3Foreign.eth.Contract(BRIDGE_VALIDATORS_ABI, foreignValidatorsAddress)
 
   logger.debug('calling foreignBridgeValidators getValidatorList()')
-  const foreignValidators = await getValidatorList(foreignValidatorsAddress, web3Foreign.eth, {
+  const foreignValidators = (await getValidatorList(foreignValidatorsAddress, web3Foreign.eth, {
     from: MONITOR_FOREIGN_START_BLOCK,
     to: foreignBlockNumber,
     logger
-  })
+  })).map(web3Foreign.utils.toChecksumAddress)
```
patitonar: While testing, I found that on some networks the addresses are returned in upper or lower case, making the comparison of the validator addresses to incorrectly detect that validators were not on the other network. To fix this, the addresses are now converted to checksum addresses

monitor/index.js
patitonar: Removed the requests related to tx costs if not needed. Only `balance` attribute will be displayed if `MONITOR_VALIDATOR_*_TX_LIMIT` is configured as zero. https://github.com/poanetwork/tokenbridge/pull/259/commits/441c944c4bc221b3550197d4eb3afb67e3692eae

monitor/validators.js
akolotov: well done

monitor/validators.js
akolotov: good catch!

monitor/validators.js
```diff
@@ -78,63 +78,86 @@ async function main(bridgeMode) {
...
-      : Web3Utils.toBN(balance)
-          .div(homeTxCost)
-          .toString(10)
-    homeVBalances[v] = {
-      balance: Web3Utils.fromWei(balance),
-      leftTx: Number(leftTx),
-      gasPrice: Number(homeGasPriceGwei)
+    if (MONITOR_VALIDATOR_HOME_TX_LIMIT) {
+      const leftTx = homeTxCost.isZero()
+        ? 999999
```
akolotov: When is it the case? Is it still needed?

akolotov:
Could you run the monitor scripts on the existing xDai bridge contracts and provide example of the output?

monitor/validators.js
patitonar: Thanks! This is not needed anymore. Fixed https://github.com/poanetwork/tokenbridge/pull/259/commits/57b3ff88e5fb10bac0a9e01c68eae5ff31cd2f08

patitonar:
This is an example output of xDai bridge for `/validators` endpoint:
```json
{
    "home": {
        "validators": {
            "0x9ad83402C19Af24F76afa1930a2B2EEC2F47A8C5": {
                "balance": "0.099801559"
            },
            "0x4D1c96B9A49C4469A0b720a22b74b034EDdFe051": {
                "balance": "0.099931099"
            },
            "0x491FC792e78CDadd7D31446Bb7bDDef876a69AD6": {
                "balance": "0.1"
            },
            "0xc073C8E5ED9Aa11CF6776C69b3e13b259Ba9F506": {
                "balance": "0.099891035"
            }
        },
        "requiredSignatures": 3
    },
    "foreign": {
        "validators": {
            "0x9ad83402C19Af24F76afa1930a2B2EEC2F47A8C5": {
                "balance": "1.792204111428",
                "leftTx": 2904,
                "gasPrice": 8
            },
            "0x4D1c96B9A49C4469A0b720a22b74b034EDdFe051": {
                "balance": "0.3276094656",
                "leftTx": 530,
                "gasPrice": 8
            },
            "0x491FC792e78CDadd7D31446Bb7bDDef876a69AD6": {
                "balance": "0.302559741248",
                "leftTx": 490,
                "gasPrice": 8
            },
            "0xc073C8E5ED9Aa11CF6776C69b3e13b259Ba9F506": {
                "balance": "0.277188330792",
                "leftTx": 449,
                "gasPrice": 8
            }
        },
        "requiredSignatures": 3
    },
    "requiredSignaturesMatch": true,
    "validatorsMatch": true,
    "lastChecked": 1578055349,
    "timeDiff": 37,
    "homeOk": true,
    "foreignOk": true,
    "ok": true
}
```

PR #276, opened 2020-01-21
Update the contract's submodule to the release 4.0.0-rc0

akolotov:
Besides updating the reference to new version of contracts (`4.0.0-rc0`) this PR will check through e2e tests that new revision does not break any existing functionality.


akolotov:
The e2e tests are failing on the `oracle-e2e` task only: most probably the issue on the side of the CircleCI servers since the issue is not permanent:

image

Please review this PR without taking into account this issue -- a separate work item will be created to address it.

Issue #100, opened 2019-06-10
Publish docker images for releases


akolotov:
It makes sense to think of an approach when an end user will pull a docker image with a bridge component instead of cloning entire git repository. In this case we could have four images on the docker hub:
  - poa/tokenbridge-oracle
  - poa/tokenbridge-ui
  - poa/tokenbridge-monitor
  - poa/tokenbrdige-deployment

The CI process could produce arbitrary images, e.g. `poa/tokenbridge-oracle:<branch-name>` or `poa/tokenbridge-oracle:<commit-hash>`. As soon as the new release is tagged, the image produced as part of the latest PR will be re-tagged like `poa/tokenbridge-oracle:<release-tag>`.

This approach should reduce deployment complexity since the number of dependencies is reduced too (e.g. you don't need to install `git`, `python` and `ansible` in order to deploy the oracle on a remote system). It also will make e2e testing for deployment component possible since will reduce nesting number of docker-in-docker beddings.

akolotov:
@phahulin @ArseniiPetrovich could you please comment the approach? Thanks!

phahulin:
I think it's a good idea and will simplify things for end-users.

> As soon as the new release is tagged, the image produced as part of the latest PR will be re-tagged like poa/tokenbridge-oracle:<release-tag>.

Who will retag it? As a suggestion - I think could reserve `master` branch for releases and configure CI to automatically publish required images for PRs into master.

rzadp:
If we take a look at the oracle deployment playbook, there is a lot more happening besides building and launching the oracle container in the `jumpbox` role, like setting up remote logging for example.

So correct me if I'm wrong, would this approach change only the way `jumpbox` role works (download oracle image instead of building it), and the end-user would still need to download the repo, install ansible and run the playbook?

akolotov:
@rzadp thanks for the question.

My expectation is that playbooks will be part of `deployment` image. The process how playbooks work will be changed as so get needed parameters from the `.env` file or some `yml` file and pass them to the `deployment` container.
So, the end user will:
1. Either perform instructions manually or download very simple shell script.
2. Prepare a configuration file and run the script
3. The script pulls the `deployment` image and runs the corresponding container passing the `env` to the container or mounting `yml` file as the volume.
4. The deployment container connects to the node which is configured in the configuration file, change every thing what is needed to run the oracle on the system - it will includes pulling the oracle/UI/monitor container to the node.

What do you think about this?

[...]

PR #280, opened 2020-01-27
Use oracle docker hub image for oracle deployment

### LINKED ISSUE #268: Use the images from the docker hub for the oracle deployment
Starting from `1.2.0-rc0` the docker images for the oracles are being published as soon as a new release is tagged. Consider to modify the deployment process to use the image from the docker hub instead of building the image on the target system.

patitonar:
Closes #268 

As part of the oracle deployment now the repo is not copied to the host, only the necessary `docker-compose*` files and the docker image is pulled instead of built in the host.

In the ultimate e2e tests, a local image is built from the current implementation of the oracle and it is used instead of the one published in docker hub

### DIFF
--- a/deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
+++ b/deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
@@ -7,7 +7,7 @@
     shell: service poabridge stop
 
   - name: Build current oracle image
-    shell: docker build -t oracle:ultimate-testing --file oracle/Dockerfile .
+    shell: docker build -t oracle:ultimate-testing --file oracle/Dockerfile --build-arg DOT_ENV_PATH=e2e-commons/components-envs/oracle.env .
     delegate_to: 127.0.0.1
     become: false
     args:
--- a/deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
+++ b/deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
@@ -11,6 +11,8 @@
       - docker-compose
       - docker-compose-transfer
       - docker-compose-erc-native
+    loop_control:
+      loop_var: file
 
   - name: start the service
     shell: service poabridge start
--- a/deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
+++ b/deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
@@ -7,7 +7,7 @@
     shell: service poabridge stop
 
   - name: Build current oracle image
-    shell: docker build -t oracle:ultimate-testing --file oracle/Dockerfile --build-arg DOT_ENV_PATH=e2e-commons/components-envs/oracle.env .
+    shell: docker build -t oracle:ultimate-testing --file oracle/Dockerfile .
     delegate_to: 127.0.0.1
     become: false
     args:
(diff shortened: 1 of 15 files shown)


deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
```diff
@@ -6,11 +6,30 @@
   - name: stop the service
     shell: service poabridge stop
 
+  - name: Build current oracle image
+    shell: docker build -t oracle:ultimate-testing --file oracle/Dockerfile --build-arg DOT_ENV_PATH=e2e-commons/components-envs/oracle.env .
```
akolotov: Do you think it is possible in the future to have the step to build the image be part of a separate job in `.circleci/config.yml`? So, overall process could be similar to:
1. Build the docker image
2. Use the docker image in tests
3. Publish the built docker image on the docker hub as intermediate image (the tag of image could be equal the git commit id)
4. If the release is tagged the built docker image will be published with the tag equals to the release version.

The idea is to have exactly the same docker image on the docker hub repo which was used for testing after the git push/merge event.

What do you think?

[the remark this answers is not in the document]
patitonar: The `.env` file is not needed for the oracle docker image itself. Right now it is used for the e2e tests.
I updated the logic in `e2e-commons/docker-compose.yml` so the .env is no longer needed in the Dockerfile https://github.com/poanetwork/tokenbridge/pull/280/commits/49291974d63068aa5eb1329d11862b53e4527955

patitonar: I think it is a good idea. It seems to be possible, here is some documentation on how to build and publish the docker images to docker hub as steps in circleci config https://circleci.com/blog/using-circleci-workflows-to-replicate-docker-hub-automated-builds/
We could open a new issue to address it

deployment/roles/oracle/tasks/jumpbox.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- name: Build the containers
-  shell: docker-compose build
+- name: Pull the containers images
+  shell: docker-compose pull
```
akolotov: Please double check. Most probably you need to remove the container explicitly here:
https://github.com/poanetwork/tokenbridge/blob/49291974d63068aa5eb1329d11862b53e4527955/deployment/roles/oracle/tasks/post_config.yml#L4
here
https://github.com/poanetwork/tokenbridge/blob/49291974d63068aa5eb1329d11862b53e4527955/deployment/roles/oracle/tasks/post_config.yml#L21
and here
https://github.com/poanetwork/tokenbridge/blob/49291974d63068aa5eb1329d11862b53e4527955/deployment/roles/oracle/tasks/post_config.yml#L32

otherwise you will get them exited but not deleted after the deployment process finish:
```
abf7dbb1e5ae        poanetwork/tokenbridge-oracle:latest   "node scripts/initia…"   43 seconds ago       Exited (0) 24 seconds ago                       oracle_bridge_affirmation_run_43b336ccf3ed
b875ee449858        poanetwork/tokenbridge-oracle:latest   "node scripts/privat…"   About a minute ago   Exited (0) 44 seconds ago                       oracle_bridge_affirmation_run_a23f8fbc6a45
```

patitonar: Good catch! I was able to reproduce the issue and fix it https://github.com/poanetwork/tokenbridge/pull/280/commits/345365b8f43e3c6cb2c58076a876a3271ee7f7bf

deployment-e2e/molecule/ultimate-commons/oracle-docker-compose.yml
akolotov: In fact, we have an issue for this: https://github.com/poanetwork/tokenbridge/issues/100

PR #287, opened 2020-02-03
Displaying chai balance in monitor

k1rill-fedoseev:


monitor/getBalances.js
```diff
@@ -143,6 +159,10 @@ async function main(bridgeMode) {
       }
     }
 
+    if (displayChaiToken) {
+      foreign.investedErc20Balance = Web3Utils.fromWei(investedAmountInDai)
```
akolotov: Is it possible to display a potential interests accumulated till this moment?

monitor/getBalances.js
```diff
@@ -103,6 +105,18 @@ async function main(bridgeMode) {
       logger.debug('Methods for half duplex token are not present')
     }
 
+    try {
+      logger.debug('calling foreignBridge.methods.chaiToken')
+      const chaiToken = await foreignBridge.methods.chaiToken().call()
+      if (chaiToken !== ZERO_ADDRESS) {
```
akolotov: does it make sense to also check `isChaiTokenEnabled`?

monitor/getBalances.js
```diff
@@ -103,6 +105,18 @@ async function main(bridgeMode) {
...
+        displayChaiToken = true
+        logger.debug('calling foreignBridge.methods.investedAmountInDai')
+        investedAmountInDai = await foreignBridge.methods.investedAmountInDai().call()
+      }
+    } catch (e) {
+      logger.debug('Methods for chai token are not present')
```
akolotov: If you are going to use `isChaiTokenEnabled` as suggested above you need also report something like "support of Chai token is disabled"

monitor/getBalances.js
patitonar: I agree I think we should use `isChaiTokenEnabled` instead of `chaiToken` since the address will be defined after migrating the contracts but could not be enabled yet.

monitor/getBalances.js
k1rill-fedoseev: Yes, used `dsrBalance() - investedAmountInDai()` for that.

monitor/getBalances.js
k1rill-fedoseev: Added logging message for disabled chai.

PR #294, opened 2020-02-06
Fix some of the possibles failure reasons in e2e tests

k1rill-fedoseev:
Seems that the usage of 3 validators in e2e tests (merged in #264) introduced several bugs to the testing workflow.

First, there are pieces of code in the tests, that are waiting for several blocks to be mined. It can happen so, that a couple of blocks won't be enough for validators to submit all required signatures. And since oracles are configured to wait 1 block confirmation, this leads to a deadlock.

Second, `isValidatorDuty` in `BridgeValidators.sol` can be called by several validators in different blocks. In case of all validators are unlucky, nobody will call `swapTokens` and the corresponding test will fail.

The following changes should fixed above problems:
- Moved blocks generation script in docker. Used this script in `oracle-e2e`, `monitor-e2e`
- Mocked `BridgeValidators.sol` for `oracle-e2e`


akolotov:
@k1rill-fedoseev it still fails

k1rill-fedoseev:
Yes, I see.
But for now, I don't know how the changes from this PR could broke this tests.

akolotov:
> But for now, I don't know how the changes from this PR could broke this tests.

Can you treat this errors slightly different: what extra improvements could be added in tests to get rid of these errors as well?

PR #302, opened 2020-03-05
Correct handling of Chai to Dai swaps in oracle and monitor components

k1rill-fedoseev:
Since now, the Chai related operations are emitting `TokensSwapped` events for swaps between Chai and Dai tokens. It is necessary to update some code for that.

* Updated monitor, so that it correctly handles `TokensSwapped` events related to Chai swaps.
* Added e2e tests for oracle to confirm that it does not handle any mint `Transfer` events when swapping Chai to Dai.

The PR relies on the changed is being prepared under https://github.com/poanetwork/tokenbridge-contracts/pull/380.


monitor/utils/events.js
```diff
@@ -112,7 +112,26 @@ async function main(mode) {
...
 
       // Get transfer events for each previous erc20
-      const uniqueTokenAddresses = [...new Set(bridgeTokensSwappedEvents.map(e => e.returnValues.from))]
+      const uniqueTokenAddressesSet = new Set(bridgeTokensSwappedEvents.map(e => e.returnValues.from))
+
+      // Exclude chai token from previous erc20
+      try {
+        logger.debug('calling foreignBridge.chaiToken()')
+        const chaiToken = await foreignBridge.methods.chaiToken().call()
+        uniqueTokenAddressesSet.delete(chaiToken)
```
akolotov: Does it make sense to filter out also by specific method signature or some additional check? Imagine that yesterday the chai token address was set but today we set it to zero. It means that this logic will work incorrectly. The same is applicable for getting `erc20token()` - the address could be changed to another token (in the case similar to the migration Sai->Dai).

k1rill-fedoseev: The Chai address is hardcoded in our contracts, so we cannot unset this address, nothing can break here.
As for `erc20token()`, another possible option is to get Dai token address from the Chai contract, where it is hardcoded. In this case, the changes to `erc20token()` in the bridge won't change filtering behaviour in the monitor.
Can I do it like this?

akolotov: Ok, I see what you mean -- changing of the hardcoded values contract equals changing of the contract logic that needs to be reflected in the monitor.  So, let's keep it as you suggested initially.

PR #304, opened 2020-03-20
Allow multiple bridges for token

varasev:
See https://github.com/poanetwork/tokenbridge-contracts/pull/387.


akolotov:
Closing this PR since another type of mutli-pegged token will be implemented as so there will be no need to change the oracle side: https://github.com/poanetwork/tokenbridge-contracts/pull/389

PR #322, opened 2020-05-06
Cache monitor RPC calls

### LINKED ISSUE #320: Optimize number of the monitor RPC calls with usage a cache
Currently the monitor consist of several workers that are run one after another:
https://github.com/poanetwork/tokenbridge/blob/5b4b01a79bb855e2caf78675cda0c5aa51024501/monitor/package.json#L7
The first two workers `checkWorker.js` and `checkWorker2.js` uses `utils/event.js` to get events from the bridge contracts. In both invocations the same JSON RPC call is used to collect events so both workers handle the same data.

In order to not spend much resources to re-work the monitor significantly consider to implement a file cache for `utils/event.js`. If the cache exists it will be used as source of the data. Creation of the cache will be managed by a new parameter in the configuration file like `MONITOR_CACHE_EVENTS`. If the parameter is `true` the cache will be created after the RPC call to get contract events.

Since the docker environment is used in the production to run the monitor workers 
https://github.com/poanetwork/tokenbridge/blob/5b4b01a79bb855e2caf78675cda0c5aa51024501/monitor/scripts/getBridgeStats.sh#L7
no extra changes are required to handle the cache between two consequent calls of the monitor scripts by cron. The cache will be stored in the container so it will be available for consecutive workers invocations but the container's file system will not persist between two cron's awakenings.

With this changes the expectation is to reduce twice amount of data received from RPC nodes by the monitor during the day in two times. For the POA bridge (exists for two years) it is very crucial.

k1rill-fedoseev:
Resolves #320


monitor/utils/events.js
```diff
@@ -16,12 +16,14 @@ const {
...
 const { filterTransferBeforeES } = require('./tokenUtils')
+const { writeFile, readCacheFile } = require('./file')
 
 const {
   COMMON_HOME_RPC_URL,
   COMMON_FOREIGN_RPC_URL,
   COMMON_HOME_BRIDGE_ADDRESS,
-  COMMON_FOREIGN_BRIDGE_ADDRESS
+  COMMON_FOREIGN_BRIDGE_ADDRESS,
+  MONITOR_CACHE_EVENTS
```
akolotov, line 26: Will it still work if the variable is not defined at all?

k1rill-fedoseev: Yes, it will. In this case `MONITOR_CACHE_EVENTS` will be equal to `undefined`.

akolotov:
Running changes on the real bridges data. So far, the POA bridge is on staging.

monitor/utils/events.js
```diff
@@ -34,7 +36,17 @@ const web3Foreign = new Web3(foreignProvider)
 
 const { getBlockNumber } = require('./contract')
 
+const cacheFilePath = './cachedEvents.json'
```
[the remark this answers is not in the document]
k1rill-fedoseev: No, the only thing to be aware of, is that this file is not located under `/mono/monitor/responses`, so that docker won't keep it between different runs

PR #328, opened 2020-05-11
Add support for STAKE token ERC677-to-ERC677 through AMB mode in Bridge UI

patitonar:
Closes #318 
Closes #316 

This PR provides:
- Correct functionality to bridge STAKE tokens on both directions
- Correct information displayed in the main page
- Correct information displayed in the statistic page
- Fixed typo in the statistic page

Pending:
- Information in the Events and Status page will be addressed in a different PR

Question:
- Should we create an issue to update ui-e2e tests to include tests for this mode? Should we also add an ultimate e2e test for this mode?

There is a deployed UI with tests contracts in Sokol-Kovan: https://stake-demo-5bffb.web.app/

This is how the statistics page looks:

statistics

Tokens Withdrawals, Tokens Deposits, Total Bridged, and Users are calculated by listening to `TokensBridged` event in home and foreign mediator contracts.

Withdrawal Fees are calculated by listening to `BridgeTokenRewardAdded` event in the home block reward contract.

Home and Foreign tokens amount are calculated by calling `totalSupply()` of each token which is requested to also display it on the main page.

Looking at the blockscout and etherscan APIs there are no methods to get data related to events generated by custom contracts that can help us to avoid querying that information directly to the blockchain. These APIs will also be re-evaluated when working on the Events and Status page.

If we want to avoid querying and perform the calculation to generate the statistics in the UI, one option could be to have a custom backend service that can provide this data directly. It should keep track of these events and provide an API so the UI can request the calculated values to show them without any other calculation. In this direction, we can investigate The graph protocol or creating our own custom solution.

PR #332, opened 2020-05-13
Display latest operations in UI status page for mediators

### LINKED ISSUE #319: Don't display validators in UI in case of bridging through AMB
When a new UI to bridge STAKE tokens through AMB is implemented it will be necessary to change the content of the status tab since the validators accounts of AMB is not relevant information for the AMB extension.

Consider to display information when the last relay operation completed. This will clearly characterise the status of the bridge: `last ETH->xDai transfer happened 2 hours ago`. Probably it makes sense to consider highlighting by different colors event of different age: “2 hours ago” by green, “2 days ago” by yellow but “1 week ago” by red.

patitonar:
Closes #319 

Built on top of #328 

The events page changes will be addressed in a separate PR.

To better review the changes, look at these commits:
- https://github.com/poanetwork/tokenbridge/commit/b093cf308ac04b7ad9cfb0f0f977c2ad809b35e4
- https://github.com/poanetwork/tokenbridge/commit/da77ffa0acd98fdcae5ff50af8c774479cbbfbc7

This is how the status page looks for STAKE mediators. The information is calculated looking for the timestamp of the block of the last `TokensBridged` event that was already requested to calculate the statistics information, so no new RPC calls were added.
statusUpdated

I used `date-fns` library to generate the label with the time distance. [Here is the documentation](https://date-fns.org/v2.13.0/docs/formatDistanceStrict ) of how the text will be updated as minutes/hours/days are increased.

I considered using different colors for the labels as suggested in the issue, but after some tests, I think it's better to only use the existing color of the UI because it looks better. So the current implementation only use the existing color.

Here are some examples of using different colors based on the time from the last operation (I tried different variations of each color). 
statusColor
statusColor2


patitonar:
Fixed the merge conflicts

PR #333, opened 2020-05-14
Add confirmRelay script

### LINKED ISSUE #326: Script to confirm a particular transaction
We have [a stuck transaction](https://etherscan.io/tx/0x764c63b2bc22fa31407c289bfc0955eb84871819e04ac18139a29a09c7b2a8a5) on the xDai bridge which could be resolved by reseting the block number on the oracle's side. But such reset is hard to implement due to intensive load for the RPC node that is caused by this action: lots of `Transfer()` needs to be re-checked that they were already confirmed.

Consider to implement an add-on to the oracle software (a script) that will allow to confirm only particular transaction:
```
confirmRelay.js <tx id>
```
this script must work in the same environment as the oracle so everything it is needed to get information about the bridge mode and the bridge contracts could be received from `.env` file. 
For this particular moment, the script must support only `erc-to-native` and the direction from Foreign to Home.

k1rill-fedoseev:
Closes #326 

Script can be run on the already working oracle, using:
```bash
docker-compose -f ./oracle/docker-compose-erc-native.yml exec \
bridge_senderforeign \
yarn confirm:transfer 0x764c63b2bc22fa31407c289bfc0955eb84871819e04ac18139a29a09c7b2a8a5
```
`bridge_senderforeign` can be replaced by any other component of the oracle which has a direct connection to `rabbitmq` service

### DIFF
--- a/oracle/src/confirmRelay.js
+++ b/oracle/src/confirmRelay.js
@@ -8,6 +8,7 @@ const { getNonce, getChainId, getEventsFromTx } = require('./tx/web3')
 const { sendTx } = require('./tx/sendTx')
 const { checkHTTPS, watchdog, syncForEach, addExtraGas } = require('./utils/utils')
 const { EXIT_CODES, EXTRA_GAS_PERCENTAGE } = require('./utils/constants')
+
 const { ORACLE_VALIDATOR_ADDRESS, ORACLE_VALIDATOR_ADDRESS_PRIVATE_KEY, ORACLE_ALLOW_HTTP_FOR_RPC } = process.env
 
 if (process.argv.length < 5) {
--- a/oracle/src/confirmRelay.js
+++ b/oracle/src/confirmRelay.js
@@ -1,0 +1,155 @@
+require('../env')
+const path = require('path')
+const { connectWatcherToQueue, connection } = require('./services/amqpClient')
+const logger = require('./services/logger')
+const rpcUrlsManager = require('./services/getRpcUrlsManager')
+const { getEventsFromTx } = require('./tx/web3')
+const { checkHTTPS, watchdog } = require('./utils/utils')
+const { EXIT_CODES } = require('./utils/constants')
+
+if (process.argv.length < 5) {
+  logger.error('Please check the number of arguments, transaction hash is not present')
+  process.exit(EXIT_CODES.GENERAL_ERROR)
+}
+
+const config = require(path.join('../config/', process.argv[2]))
+const txHash = process.argv[4]
+
+const processSignatureRequests = require('./events/processSignatureRequests')(config)
+const processCollectedSignatures = require('./events/processCollectedSignatures')(config)
+const processAffirmationRequests = require('./events/processAffirmationRequests')(config)
+const processTransfers = require('./events/processTransfers')(config)
+const processAMBSignatureRequests = require('./events/processAMBSignatureRequests')(config)
+const processAMBCollectedSignatures = require('./events/processAMBCollectedSignatures')(config)
+const processAMBAffirmationRequests = require('./events/processAMBAffirmationRequests')(config)
+
+const { getTokensState } = require('./utils/tokenState')
+
+const web3Instance = config.web3
+const bridgeContract = new web3Instance.eth.Contract(config.bridgeAbi, config.bridgeContractAddress)
+let { eventContractAddress } = config
+let eventContract = new web3Instance.eth.Contract(config.eventAbi, eventContractAddress)
+
+async function initialize() {
+  try {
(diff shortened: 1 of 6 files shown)


oracle/src/confirmRelay.js
```diff
@@ -0,0 +1,155 @@
...
+
+async function checkConditions() {
+  let state
+  switch (config.id) {
+    case 'erc-native-transfer':
+      logger.debug('Getting token address to listen Transfer events')
+      state = await getTokensState(bridgeContract, logger)
+      updateEventContract(state.bridgeableTokenAddress)
+      break
+    case 'erc-native-half-duplex-transfer':
```
akolotov, line 109: Do we still need to support a HD token transfer?

oracle/src/confirmRelay.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+}
+
+async function checkConditions() {
+  let state
+  switch (config.id) {
+    case 'erc-native-transfer':
+      logger.debug('Getting token address to listen Transfer events')
+      state = await getTokensState(bridgeContract, logger)
+      updateEventContract(state.bridgeableTokenAddress)
```
akolotov, line 107: Do we need this? As far as I recall it was used to handle a case when `bridgeableTokenAddress` is not defined yet. Please double check if we can simplify the code here.

oracle/src/confirmRelay.js
```diff
@@ -0,0 +1,155 @@
...
+      }, config.pollingInterval)
+    }
+  } catch (e) {
+    logger.error(e)
+  }
+}
+
+function processEvents(events) {
+  switch (config.id) {
+    case 'native-erc-signature-request':
```
akolotov, line 74: Please clarify why do we need to process Signature Requests and Collected Signatures events here?

oracle/src/confirmRelay.js
```diff
@@ -0,0 +1,155 @@
...
+      filter: config.eventFilter
+    })
+    logger.info(`Found ${events.length} ${config.event} events`)
+
+    if (events.length) {
+      const job = await processEvents(events)
+      logger.info('Transactions to send:', job.length)
+
+      if (job.length) {
+        await sendToQueue(job)
```
akolotov, line 143: Does it mean that we need to have the oracle up and running? Could you estimate how much it could take to have two options: the oracle has been run before `confirmRelay.js` is invoked and the oracle has been stopped. It will be useful in the future to have a script which could operate with the signer private key without need to deploy and run entire oracle.

oracle/src/confirmRelay.js
k1rill-fedoseev: No, I removed this piece of code in b4fedc3.

oracle/src/confirmRelay.js
k1rill-fedoseev: Yes, you are right, it will work fine without this. I simplified it in b4fedc3.

oracle/src/confirmRelay.js
k1rill-fedoseev: I decided to keep almost all events and processing logic from `oracle/src/watcher.js`. So it should be possible to process any individual transaction using new script, if it contains some unprocessed events (including Signature requests / Collected signatures events).

oracle/src/confirmRelay.js
k1rill-fedoseev: I have extended the script in the following way:
* if `ORACLE_QUEUE_URL` is available, the script will use running oracle's environment to send a tx (it will add a new job in a queue).
* if `ORACLE_QUEUE_URL` is not available, the script will send a tx by itself using provided private key.

For the second option, the script can be simply run by `docker run --env-file oracle/.env poanetwork/tokenbridge-oracle:latest yarn confirm:transfer 0x76..a5`

oracle/src/confirmRelay.js
akolotov: I understood this. And my worry is that we don't lose usability by trying to achieve universality.  Please confirm that for the bad case the person who runs this functionality will get a descriptive reason why the action cannot be completed (e.g. that confirmation from this validator has been already registered in the contract).

k1rill-fedoseev: This reasons are present in the handling logic in `oracle/src/events/*`, so everything should be ok.
For example, for some already handled transaction I am getting the following message while trying to process `CollectedSignatures` event:
`[1589897350889] INFO  (watcher-erc-native-collected-signatures): Validator not responsible for relaying CollectedSignatures 0xf0d89e4cd07981d528c2915b4f7ba3878b0aa2b849dba21f952fc815d1bca5e5`

akolotov: thanks for explanation

PR #334, opened 2020-05-16
Health field in the monitor reports 

akolotov:
Set of improvements added to the monitor:
- the running scripts sets `health` to `false` for monitor reports before running the monitor workers for a particular bridge configuration
- the running scripts does not run the workers if they do not stop by some reason for a particular bridge configuration
- the successful execution of workers will set `health` in the report to `true`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,4 +49,5 @@ __pycache__
 
 #monitor
 monitor/responses/*
+monitor/configs/*.env
 !monitor/.gitkeep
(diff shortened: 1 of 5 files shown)


akolotov:
@patitonar please take a look one more time. I have modified the logic under https://github.com/poanetwork/tokenbridge/pull/334/commits/01af787bd9b73b6e38940d7733d49ae3e8bc7fd0. It gets rid of the gap between the state 'health:false' and the state `health:true`.

PR #341, opened 2020-05-22
Update UI for metamask breaking changes

### LINKED ISSUE #229: Support metamask breaking changes in new API
Breaking changes are listed in this post https://medium.com/metamask/breaking-changes-to-the-metamask-inpage-provider-b4dde069dd0a

The only change that seems to affect the Bridge-UI component is: 
- Stop reloading the page on network change

A [possible solution](https://medium.com/metamask/no-longer-reloading-pages-on-network-change-fbf041942b44) is to emulate the current behavior by subscribing to `chainChanged` event and trigger a reload

```javascript
ethereum.on('chainChanged', () => {
  document.location.reload()
})
```
when we detect the `ethereum` injected object 
https://github.com/poanetwork/tokenbridge/blob/8b010f887dbe9c369e3f30cf9ea0d3a59c83fe25/ui/src/stores/utils/web3.js#L29-L41

patitonar:
Closes #229 

The new API is not available yet. This comment is the latest update on these changes https://github.com/MetaMask/metamask-extension/issues/8077#issuecomment-622191567

Based on the details of the changes:
- We are already using `window.ethereum`
- We are using our own `web3` version and not the injected one. The injected one is only used to get the provider for other wallets that don't inject `window.ethereum`.
- We are not using the event `chainIdChanged` but we are using the automatic page reload on network change.

So in order for the UI to continue working with metamask in the future without issues the only change needed is to manually trigger the page reload when on a network change by listening to the new event `chainChanged` that will be later introduced in the next versions.

New API docs: https://docs.metamask.io/guide/ethereum-provider.html#methods-new-api


patitonar:
>Can you elaborate this a little bit more? I am not completely sure that I got the last sentence.

Sure, let me add more details.

We are using our own imported Web3 version
https://github.com/poanetwork/tokenbridge/blob/2f70fa61aae6c6a9e7e011e8bcf412e73b48074b/ui/src/stores/utils/web3.js#L1
In the current version of Metamask/Nifty wallet the correct way to get access to the network the user is connected/user accounts is by using the injected `window.ethereum` object
https://github.com/poanetwork/tokenbridge/blob/2f70fa61aae6c6a9e7e011e8bcf412e73b48074b/ui/src/stores/utils/web3.js#L25-L33

But in previous versions, before the `window.ethereum` was introduced, the `windows.web3` object was used
https://github.com/poanetwork/tokenbridge/blob/2f70fa61aae6c6a9e7e011e8bcf412e73b48074b/ui/src/stores/utils/web3.js#L24
https://github.com/poanetwork/tokenbridge/blob/2f70fa61aae6c6a9e7e011e8bcf412e73b48074b/ui/src/stores/utils/web3.js#L47-L48

So it could be the case where the wallet used is not compliant with the new `window.ethereum` and continued to use the `windows.web3` object. In that case, if the `ethereum` object is not present, our code will try to use the `web3` object to get access to the network selected in the wallet

akolotov:
So, in both cases we do not rely on the web3 functionality provided by the the wallet, do we? We just use the wallet as the source of a Web3 provider in order to communicate with the provider through web3 provisioned with the TokenBridge UI.

patitonar:
>So, in both cases we do not rely on the web3 functionality provided by the the wallet, do we? We just use the wallet as the source of a Web3 provider in order to communicate with the provider through web3 provisioned with the TokenBridge UI.

That's correct

PR #347, opened 2020-05-29
Add UI e2e and ultimate tests for stake mediators and fix transfer finalization detection

### LINKED ISSUE #329: Add ui-e2e and ultimate e2e test for STAKE token
Changes were introduced in https://github.com/poanetwork/tokenbridge/pull/328

patitonar:
Closes #329 
Built on top of #346 

This PR adds:
- UI e2e tests for stake mediators
- Ultimate test for stake medaitors
- Fix transfer finalization detection when using new version of AMB

While adding the e2e tests, I found that the UI wasn't detecting correctly the transfer finalization. The reason was that a new messageId was introduced when adding support for multiple messages in one transaction in AMB. That's why a fix was needed to make it also work for this new version. The fix was applied mainly in this commit https://github.com/poanetwork/tokenbridge/commit/3d55d565c1a9bc56ab3c95d6f4ccd542123088f0

That's why this PR was built on top of #346  to use the logic added there to get the AMB contracts from the mediatos.


patitonar:
Updated the AMB version to `5` https://github.com/poanetwork/tokenbridge/pull/347/commits/669d057b586ad9ef1290b971a542bfa3b3ade816

I double checked, we are calling `getBridgeInterfacesVersion` on the AMB contracts on [home](https://github.com/poanetwork/tokenbridge/blob/52d02f74759e44d626020e327d79564873f65580/ui/src/stores/HomeStore.js#L679) and [foreign](https://github.com/poanetwork/tokenbridge/blob/52d02f74759e44d626020e327d79564873f65580/ui/src/stores/ForeignStore.js#L544) side.

akolotov:
> I double checked, we are calling `getBridgeInterfacesVersion` on the AMB contracts

muchas gracias

PR #348, opened 2020-06-02
Add ALM project

patitonar:
This PR adds an empty typescript React app in `alm` folder including the setup files, yarn scripts, Dockerfile and docker-compose files.


patitonar:
Fixed a typescript version conflict between alm and bw plugin that was causing the build of the plugin to fail. Also updated the `build` script of the monorepo to also build the plugin, so the CI can catch any other future errors

PR #356, opened 2020-06-12
Add ALM transaction verification detection

patitonar:
This PR adds the following features:
- Detect that validators are waiting for block confirmations for the request
- Detect that the responsible validator is waiting for block confirmations to execute the signatures on the foreign network
- Detect the submitted verifications by the validators by calling `messagesSigned` and `affirmationsSigned` methods
- Detect the message execution finalization by listening to the events `RelayedMessage` and `AffirmationCompleted`

Another change is that In the transaction information page now the form is not displayed. I removed it to save some space on the page and also it looks cleaner to me. You can go back to the form by clicking the title. If it is not so obvious, we can also add some button to go back.

You will notice some differences in the displayed information with the design from the slides:
- The `Validator` table does not include the `Age` column, this is because we don't have the transaction to display the information about it. In the future, we can add the column if we provide the functionality to discover this information.
- The `Executed by` table is also displayed for foreign to home transactions. Despite the execution not being done in a separate transaction, it helps to display which validator executed the message (by being the last to submit the signature) and to display the time since the execution including the link to the transaction. If you think displaying this is confusing, we can think of better alternatives.

The demo page was updated with these changes: https://alm-demo-e0998.web.app
**Edit:** The demo also includes latest changes from https://github.com/poanetwork/tokenbridge/pull/357

Example Home to Foreign transaction
https://alm-demo-e0998.web.app/77/0x5325279181e04916f4a3e95fc0b34c02afc0d6392e79b53fe58b2ff69f8f2779

Example Foreign to Home transaction
https://alm-demo-e0998.web.app/42/0x480c5f8abe241e52f756d9d5ea9faeb8b633d44a6441cdfc0cf194b3489e5699

The following features will be addressed in futures PRs:
- Detect Failed transactions
- Detect Pending transactions


patitonar:
>Yes, it is not obvious. Consider to add another element with a tiny text like "Check another transaction" that needs to be clicked to display the input field.

Created issue https://github.com/poanetwork/tokenbridge/issues/362 for this

>Is this since we are not working with pending transactions? So, when we implement requesting them this column could appear, right?

Actually, it is because we detect the submitted verifications by the validators by calling `messagesSigned` and `affirmationsSigned` methods so we don't have the transaction. I didn't plan to include the Age column with the implementation of getting the pending transaction. While we can get the pending transaction (and so get the transaction hash value), it is not garanteed because even if we are live watching the ALM page as the message is being processed, we could get the confirmation that the validator submitted the signature before even querying the pending transaction result (usually home networks are fast), or we can receive an empty list since it was already processed by that moment, or we could only get them for some validators. Also, in the case of an already processed message, there won't be any pending transaction in the list for this message. 

One option that you mentioned in the document could be to later (if it is needed) add a button/an option to get more details for such confirmations: when this button is pressed the ALM will try to discover the corresponding transactions.

akolotov:
In fact, providing information to the user about transaction ageing is a key feature of ALM: by having more information the user is able to understand what exactly is happening with his/her transfer request. If only the 'waiting' message appears it will say nothing.

Moreover currently you solution relies on the validator list taken from the current state of the chain. But in general the list could be changed. So, if you provide a transaction that was confirmed but the validator set was changed after that, the ALM will think that validators didn't finished their work. Am I right?

What do you think about the following approach (I will try to present by using the process of handling the `UserRequestForSignature` event):
1. detect that the corresponding `UserRequestForSignature` emitted by the transaction specified by the user.
2. get the list of validators for the current state.
3. get events from the validator contract to actualize the list of validators. The logs should be get starting from the block where the transaction emitted `UserRequestForSignature` was mined.
4. check `messagesSigned` for every validator to display the status of the confirmation
5. check if the request was executed by requesting `relayedMessages` on the foreign side
6. by using known validator accounts, get the transactions they sent to confirm the request. Use these transactions to set the age for the confirmations and status (e.g. failed)
7. if the validator transaction is not found, request the pending transactions list from this validator. Reflect the status of the confirmation (e.g. pending or waiting).
8. by using the validator address that was needed to send signatures, get the transaction with these signatures to set the age and status (failed)
9. if the transaction is not found, request the pending transactions list from this validator. Reflect the status of the confirmation (e.g. pending or waiting)

For 6 and 7 the starting block to limit the transaction lookup equals the block from no. 3.
For 8 and 9 the starting block could be assumed by a procedure we discussed before.

Note that 6-9 are executed only after attempts to get information from the contract state. It helps to user get information on earlier stages of requesting ALM and wait for extended info after that.

Don't think that adding one more button will improve the user experience.

Please share your view on my suggestions.

patitonar:
>Moreover currently you solution relies on the validator list taken from the current state of the chain. But in general the list could be changed. So, if you provide a transaction that was confirmed but the validator set was changed after that, the ALM will think that validators didn't finished their work. Am I right?

That's Correct! Right now it is using the current list of validators instead of the list of validators at the moment of the transaction.

>2. get the list of validators for the current state.
>3. get events from the validator contract to actualize the list of validators. The logs should be get starting from the block where the transaction emitted UserRequestForSignature was mined.

For this, I would suggest [specifying the block number](https://github.com/ethereum/web3.js/issues/1287) when calling the method `validatorList` and `requiredSignatures` so we can get the information at that block. We can use the block number from 1 and call the validator methods from home or foreign to get the list and required signatures based where the transaction was originated so we can use the block number.

>5. check if the request was executed by requesting relayedMessages on the foreign side

Right now, instead of requesting `relayedMessages` on the foreign side, after we validate that the tx has the signatures by calling `messagesSigned`, we are listening to the event `CollectedSignatures` from the block of the transaction 1. By getting this event, we have the information of who is the validator responsible for relying the transaction on the foreign side (to display it and also to later request the failed and pending transaction list), and if he is waiting for block confirmations. So I think we can leave it as it is now. **What do you think?**

>6. by using known validator accounts, get the transactions they sent to confirm the request. Use these transactions to set the age for the confirmations and status (e.g. failed)

This is partially done in https://github.com/poanetwork/tokenbridge/pull/357 but only for failed transactions and instead of using the block number, it uses the timestamp of the block which blockscout allows to set to filter the list and also would work for transactions from home and from foreign (otherwise in tx from foreign we will need to calculate the block on home based on the timestamp). Also, the Age is not displayed.

> 7. if the validator transaction is not found, request the pending transactions list from this validator. Reflect the status of the confirmation (e.g. pending or waiting).

This is implemented in https://github.com/poanetwork/tokenbridge/pull/363

> 8. by using the validator address that was needed to send signatures, get the transaction with these signatures to set the age and status (failed)

Right now to get the success execution we are getting the `RelayedMessage` and `AffirmationCompleted` and displaying the Age. If it is not found, it will use blockscout to get the transactions looking for the failed transaction which is implemented in https://github.com/poanetwork/tokenbridge/pull/357 **Do you think we need to change it to get the successful transaction also as part of the transactions requested to blockscout?**

>9. if the transaction is not found, request the pending transactions list from this validator. Reflect the status of the confirmation (e.g. pending or waiting)

This is implemented in https://github.com/poanetwork/tokenbridge/pull/363

So, unless I'm missing something, and waiting for the agreement on some of the details I asked, there are 2 improvements:
- Get the correct list of validators and required signatures for the block of the transaction
- Display the Age for the validator signatures. To get the transaction for the successful signatures, request the list of transactions for each validator using blockscout as we do for the failed transaction

**Do you think we can merge this PR and add the improvements as separates issues?** It will help to unlock the others PR

akolotov:
> For this, I would suggest specifying the block number when calling the method validatorList and requiredSignatures so we can get the information at that block. We can use the block number from 1 and call the validator methods from home or foreign to get the list and required signatures based where the transaction was originated so we can use the block number.

Right. But it could not work if the node was not configured as an archive node. For example, you could not get very old state by using Infura. That's why I suggested to use events to get an actual state.

> Right now, instead of requesting relayedMessages on the foreign side, after we validate that the tx has the signatures by calling messagesSigned, we are listening to the event CollectedSignatures from the block of the transaction 1. By getting this event, we have the information of who is the validator responsible for relying the transaction on the foreign side (to display it and also to later request the failed and pending transaction list), and if he is waiting for block confirmations. So I think we can leave it as it is now. What do you think?

That's OK for me.

> Do you think we need to change it to get the successful transaction also as part of the transactions requested to blockscout?

No, that's OK.

> Do you think we can merge this PR and add the improvements as separates issues? It will help to unlock the others PR

Yes, let's do as you suggested. Please open an issue and prepare another PR. Thank you.

patitonar:
> Right. But it could not work if the node was not configured as an archive node. For example, you could not get very old state by using Infura. That's why I suggested to use events to get an actual state.

Thanks, I missed that. Let's do it as you suggested then.

> Get the correct list of validators and required signatures for the block of the transaction

Created https://github.com/poanetwork/tokenbridge/issues/365

> Display the Age for the validator signatures. To get the transaction for the successful signatures, request the list of transactions for each validator using blockscout as we do for the failed transaction

Created https://github.com/poanetwork/tokenbridge/issues/366

PR #367, opened 2020-06-23
Build the list of validators and required signatures at the moment of the transaction

### LINKED ISSUE #365: Get the correct list of validators and required signatures at the moment of the transaction
The steps could be:
1. Get the list of validators for the current state.
2. Get events from the validator contract to actualize the list of validators. The logs should be get starting from the block where the transaction was mined

patitonar:
Closes #365


akolotov:
Do you have another deployment with the current state of the application for review?

patitonar:
I updated the demo to include the changes from `develop` branch and this PR: https://alm-demo-e0998.web.app/

alm/src/utils/contract.ts
```diff
@@ -7,12 +7,46 @@ export const getRequiredBlockConfirmations = async (contract: Contract) => {
 
 export const getValidatorAddress = (contract: Contract) => contract.methods.validatorContract().call()
 
-export const getRequiredSignatures = async (contract: Contract) => {
-  const requiredSignatures = await contract.methods.requiredSignatures().call()
+export const getRequiredSignatures = async (contract: Contract, blockNumber: number) => {
+  const events = await contract.getPastEvents('RequiredSignaturesChanged', {
+    fromBlock: 0,
```
akolotov, line 12: Consider to have this changes applied to the Ethereum Mainnet, does it make sense to start from the block 0?

Do you think it is possible to generate a snapshot of events (not only `RequiredSignaturesChanged`) for the ALM application as part of deployment process so the application will request just delta between the snapshot and current state of the blockchain? I believe it could make the application more respondent.

patitonar: > Consider to have this changes applied to the Ethereum Mainnet, does it make sense to start from the block 0?

I thought about it, one alternative is to first call `deployedAtBlock` and then use it as the `fromBlock` parameter. But in this case, I think it would take more time to get the result because of doing one request, wait for the result, and then do the second request. That's why I have chosen to use the block `0`

> Do you think it is possible to generate a snapshot of events (not only RequiredSignaturesChanged) for the ALM application as part of deployment process so the application will request just delta between the snapshot and current state of the blockchain? I believe it could make the application more respondent.

I think it is possible. We could develop a script that will get for example the events of `RequiredSignaturesChanged`, `ValidatorAdded`, and `ValidatorRemoved`. The script could run before building the application and store the results in a file that will be included in the sources (to avoid having other external services). Then the app will get the information from these snapshots and only request the delta.

The snapshots could also include the basic information of the app such as:
- Chain Ids
- Block confirmations numbers
- Validator contracts address (and for example `deployedAtBlock` value if needed) 

Having this information will save doing a lot of requests, which will increase the response of the app. I wouldn't recommend storing information about the user's transactions because it could be a lot of data, and it will affect the loading time of the app because of the size of the snapshots.

Should we create a new issue for snapshot improvement?

akolotov: > Block confirmations numbers

BTW, this number could be changed. And there is no way to track these transactions by event, isn't it?

> Should we create a new issue for snapshot improvement?

Yes, please

patitonar: >BTW, this number could be changed. And there is no way to track these transactions by event, isn't it?

The event `RequiredBlockConfirmationChanged` is emitted when the number is changed, so I think it makes sense to also create an issue to get the block confirmations similar as we are doing for required signatures using the events

akolotov: Cool! Let's create an issue and keep in mind that these events also should be in a snapshot we discussed above.

patitonar: Created the following issues:
- For Block confirmations changes https://github.com/poanetwork/tokenbridge/issues/369
- For snapshot https://github.com/poanetwork/tokenbridge/issues/370

PR #373, opened 2020-06-25
Add ALM new styles

### LINKED ISSUE #368: Update ALM UI styles similar to easy staking app

### LINKED ISSUE #372: Remember the last used chain
When "Search another transaction" button is pressed the default chain is chosen automatically. Consider to remember the last used chain.

patitonar:
Built on top of https://github.com/poanetwork/tokenbridge/pull/371

This PR adds the following changes:
- New UI styles. Closes #368 
- Remember the last used chain. Closes #372 
- Add a new definition of the ABIs used inside the project sources. https://github.com/poanetwork/tokenbridge/commit/3aa951caa4731ca08bf8fb81636bc4961bac0961

I noticed that the bundle size of the app after running the `build` command was kind of big. After inspecting it, I detected that the whole `commons` package with all the ABIs and defined method was being included in the bundle. I'm not sure why webpack (the tool used by create-react-app to generate the build) is failing at [removing the resources that are not used](https://webpack.js.org/guides/tree-shaking/) for that module, finding the cause will require some more in-depth analysis. To avoid this issue, I included the ABI used by ALM as part of the project.

Before sizes
```
531.01 KB          build/static/js/main.3b16c90d.chunk.js
438.47 KB          build/static/js/2.35b98736.chunk.js
762 B              build/static/js/runtime~main.a8a9905a.js
```

After sizes
```
419.25 KB (-19.22 KB)  build/static/js/2.dab69f3e.chunk.js
12.08 KB (-518.92 KB)  build/static/js/main.c6b2322e.chunk.js
762 B                  build/static/js/runtime~main.a8a9905a.js
```

Updated the Demo page to include these changes: https://alm-demo-e0998.web.app/

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@ submodules
 coverage
 lib
 dist
+build
--- a/alm/package.json
+++ b/alm/package.json
@@ -14,6 +14,7 @@
     "@types/react-dom": "^16.9.0",
     "@types/react-router-dom": "^5.1.5",
     "@types/styled-components": "^5.1.0",
+    "@use-it/interval": "^0.1.3",
     "customize-cra": "^1.0.0",
     "date-fns": "^2.14.0",
     "fast-memoize": "^2.5.2",
--- a/alm/public/index.html
+++ b/alm/public/index.html
@@ -25,6 +25,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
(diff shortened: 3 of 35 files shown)


patitonar:
I'm OK adding a dropdown list in the top right corner in the header, but I'm not sure it is the best option for the user. Right now it seems to me that it is clearer that the user has to select one network when searching the transaction while having the dropdown in the header may not be that clearer.

After searching the transaction, on the results page, I think the user shouldn't be able to select a new network from the dropdown, and instead, it should only show the currently selected network as it is now. What do you think?

I'll add this new functionality in a separate PR

akolotov:
OK. Probably, you are right.

PR #378, opened 2020-06-29
Extend burner wallet plugin with support of erc-to-native mediator

k1rill-fedoseev:


burner-wallet-plugin/tokenbridge-bw-exchange/src/burner-wallet/assets/BridgeableERC20Asset.ts
```diff
@@ -0,0 +1,48 @@
...
+      const Contract = this.getWeb3().eth.Contract
+      this._bridge = new Contract(MEDIATOR_ABI, this.bridgeAddress)
+    }
+    return this._bridge
+  }
+
+  async _send({ from, to, value }) {
+    if (to.toLowerCase() === this.bridgeAddress) {
+      const allowance = await this.allowance(from, to)
+      if (toBN(allowance).lt(value)) {
```
patitonar, line 34: While testing, I found that the `approve` method was never called even if the allowance was `0`. Converting `value` to BN will properly compare the two numbers and call `approve` if required
```suggestion
      if (toBN(allowance).lt(toBN(value))) {
```

k1rill-fedoseev: You are right. Thanks

burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/abis/Mediator.ts
```diff
@@ -34,5 +34,45 @@ export default [
...
+        name: '',
+        type: 'address'
+      },
+      {
+        indexed: false,
+        name: '',
+        type: 'uint256'
+      }
+    ],
+    name: 'Transfer',
```
akolotov, line 75: The mediator should not emit this event. Could you clarify why it is here?

burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/utils.ts
```diff
@@ -7,7 +7,9 @@ export const constants = {
   EXCHANGE_TIMEOUT: 300000,
   MAX_FEE: toWei('1', 'ether'),
   ESTIMATE_FEE_MESSAGE: 'Estimation takes fee charges into consideration.',
-  ZERO_ADDRESS: '0x0000000000000000000000000000000000000000'
+  ZERO_ADDRESS: '0x0000000000000000000000000000000000000000',
+  HOME_TO_FOREIGN_FEE_TYPE: '0x741ede137d0537e88e0ea0ff25b1f22d837903dbbee8980b4a06e8523247ee26',
```
akolotov, line 11: Could you add in the comment a reference to the the corresponding tokenbridge contract?

burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/abis/Mediator.ts
k1rill-fedoseev: When tokens are sent through the bridge using `relayTokens`, the method is called on the bridge contract instance: https://github.com/poanetwork/tokenbridge/blob/5ee9dea0c44cf8f2b055333886cc79f0ffc6251c/burner-wallet-plugin/tokenbridge-bw-exchange/src/burner-wallet/assets/BridgeableERC20Asset.ts#L37-L39
This call results in Transfer event being emitted from the token contract, however, when web3 parses receipt logs, it knows only information about events from the called contract. So I added this ABI here in order to further access parsed Transfer event at https://github.com/poanetwork/tokenbridge/blob/5ee9dea0c44cf8f2b055333886cc79f0ffc6251c/burner-wallet-plugin/tokenbridge-bw-exchange/src/burner-wallet/assets/BridgeableERC20Asset.ts#L43

burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/utils.ts
k1rill-fedoseev: Sure

burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/abis/Mediator.ts
akolotov: Yeah, I understood this. This is a sort of shortcut but could lead to a confusion in the future. Is this more correct to get the transaction hash and use the token contract to get the Transfer event by using the transaction receipt? Yes, it will add more code but this is more correct to reflect which contract which event emitted.

k1rill-fedoseev: Ok, agree. Since we don't use parsed event and the only thing we need is its log index, we can find a suitable log by comparing its topic with the existing `TRANSFER_TOPIC` constant.

PR #380, opened 2020-07-01
ALM docker improvements

patitonar:
This PR introduces the following changes:
- Fixes an error with web3 dependencies that caused the app to failed to compile using docker
- Allow specifying etherscan key. Example: `ALM_FOREIGN_EXPLORER_API=https://kovan.etherscan.io/api?apikey=YourApiKeyToken`
- Updated docker-compose file to bind ports. The new variable `PORT` is used for this.
- Updated Dockerfile to serve the production build of the app. This also allows reducing the size of the docker image to `~950MB`

Example of building and running with docker:
```
docker-compose build
```

And then
```
docker-compose up
```
or
```
docker run -ti -p 8080:8080 -e PORT=8080 --rm alm_alm:latest
```

### DIFF
--- a/alm/Dockerfile
+++ b/alm/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12
+FROM node:12 as alm-builder
 
 WORKDIR /mono
 COPY package.json .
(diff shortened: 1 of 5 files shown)


alm/Dockerfile
```diff
@@ -19,5 +19,11 @@ ARG DOT_ENV_PATH=./alm/.env
...
-CMD echo "To start the application run:" \
-  "yarn start"
+RUN yarn run build
+
+
+FROM node:12 as alm-production
+RUN yarn global add serve
+WORKDIR /app
+COPY --from=alm-builder /mono/alm/build .
+CMD echo "To start the application run: serve -p 8080 -s ."
```
akolotov, line 29: could you check that this form of running the application will allow to use it with the Heroku?

patitonar: Updated the CMD command and the variable used, to make it work with Heroku https://github.com/poanetwork/tokenbridge/pull/380/commits/493e16c4f8e0db6afccb7efabbbeed80947513f3

Issue #370, opened 2020-06-24
Generate a snapshot of useful information to be included in ALM


patitonar:
Develop a script that will get useful information for the app. The script could run before building the application and store the results in a file that will be included in the sources (to avoid having other external services). Then the app will get the information from these snapshots and only request the delta.

The snapshots could also include the basic information of the app such as:

- Chain Ids
- Block confirmations numbers
- Validator contracts address (and for example deployedAtBlock value if needed)
- RequiredSignaturesChanged events
- ValidatorAdded events 
- ValidatorRemoved events
- RequiredBlockConfirmationChanged events

Having this information will save doing a lot of requests, which will increase the response of the app. I wouldn't recommend storing information about the user's transactions because it could be a lot of data, and it will affect the loading time of the app because of the size of the snapshots.

PR #382, opened 2020-07-02
Add ALM snapshots

patitonar:
Closes https://github.com/poanetwork/tokenbridge/issues/370

Add a script that runs before the app is built. The script will get useful information about the bridge contracts and networks and store it in the sources of the app to reduce the number of requests needed when the user visits a page.

Another fix introduced here is to set a fixed size for the columns in the tables for validators confirmations and execution in https://github.com/poanetwork/tokenbridge/commit/072cd9dfef845d94a2aa4e5d3ee345ec0bc07a4c. Previously, the column adjusted the size based on the content provided which caused the columns to move and not be correctly aligned.

Updated the demo: https://alm-demo-e0998.web.app/


alm/package.json
```diff
@@ -30,11 +31,12 @@
     "web3-eth-contract": "1.2.7"
   },
   "scripts": {
-    "start": "./load-env.sh react-app-rewired start",
-    "build": "./load-env.sh react-app-rewired build",
+    "start": "yarn createSnapshots && ./load-env.sh react-app-rewired start",
```
akolotov, line 34: will it work with the new dockerfile which uses `serve`?

alm/src/utils/web3.ts
```diff
@@ -61,3 +62,11 @@ export const getBlock = async (web3: Web3, blockNumber: number): Promise<BlockTr
     }
     return result
   })
+
+export const getChainId = async (web3: Web3, snapshotProvider: SnapshotProvider) => {
+  let id = snapshotProvider.chainId()
```
akolotov, line 67: We don't get lots of advantages if get the chain id from the snapshot rather than from the chain. At the same time, if something goes wrong (e.g. a chain with an another chain id is available through JSON-RPC URL) getting a real value from the chain will not confuse users.

Instead I suggest to use some info (e.g. the block hash of the latest block) from the snapshot to confirm that the snapshot was taken for the same chain as is being used in the ALM app when a user works with it.

alm/src/utils/contract.ts
```diff
@@ -19,33 +33,58 @@ export const getRequiredBlockConfirmations = async (contract: Contract, blockNum
   return parseInt(blockConfirmations)
 }
 
-export const getValidatorAddress = (contract: Contract) => contract.methods.validatorContract().call()
+export const getValidatorAddress = async (contract: Contract, snapshotProvider: SnapshotProvider) => {
```
akolotov, line 36: The validator contract address could be changed later so data in the snapshot could be outdated.   There is no big advantage if we get them form the snapshot.

alm/src/services/SnapshotProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+  event: string
+}
+
+export interface Snapshot {
+  chainId: number
+  RequiredBlockConfirmationChanged: SnapshotEvent[]
+  validatorAddress: string
+  RequiredSignaturesChanged: SnapshotEvent[]
+  ValidatorAdded: SnapshotValidatorEvent[]
+  ValidatorRemoved: SnapshotValidatorEvent[]
```
akolotov, line 28: Could you estimate the growth of the snapshot if we include all transactions from the bridge validators to it? What information received from the chains by the ALM could be included in the snapshot as well?

alm/src/services/SnapshotProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+  validatorAddedEvents(fromBlock: number) {
+    return this.data.ValidatorAdded.filter(e => e.blockNumber >= fromBlock)
+  }
+
+  validatorRemovedEvents(fromBlock: number) {
+    return this.data.ValidatorRemoved.filter(e => e.blockNumber >= fromBlock)
+  }
+}
+
+export const homeSnapshotProvider = new SnapshotProvider('home')
```
akolotov, line 74: Is it worth to use the chain id as an identifier instead of `home` and `foreign`?

patitonar: This identifier is used to fill the path to find the snapshot in `data = require('../snapshots/${side}.json')` and at the moment of creating the SnapshotProvider intances the app does not have the chainId, it will get it later from the snapshot or from the blockchain if not available in the snapshot.

alm/package.json
patitonar: Yes, the script to create the snapshot is also called in the `build` script. It will generate the snapshots and include them in the sources of the app in the folder `src/snapshots` so they will be included in the code of the built app
https://github.com/poanetwork/tokenbridge/blob/2e062ff1d715229520a878a538cabdf9842e7124/alm/package.json#L35

alm/src/utils/contract.ts
patitonar: In which scenario do you think the validator contract address could be changed? Validator contracts are usually a proxy too and the bridge contract does not provide a method for such functionality

alm/src/utils/web3.ts
patitonar: I'm not sure I fully understood your suggestion and what are we trying to achieve with it.
The advantage of getting the chainId from the snapshot is to avoid doing 2 requests to get the chainId from the chain (one for home and one for foreign).

The snapshot script creates two files, one for home network (in `src/snapshots/home.json`) and one for foreign network (in `src/snapshots/foreign.json`) with the information collected in the latests available block using the env vars for the RPC URLS and the bridge contract addresses. So all the information stored in the snapshot will be related to the same chain.

So when the user visits a page, the chainId will be read from the snapshot instead of sending a request to the RPC URL. If the RPC URL is wrong and now it is providing data from a different chain most probably the information fetched from the RPC will be all wrong, the address of the bridge contract provided by env var could not exist in that chain, it won't be able to get the receipt, the correct addresses, list of validators, etc, even if the snapshot is not present.

alm/src/services/SnapshotProvider.ts
patitonar: This is how the snapshot looks like for the sokol - kovan amb bridge for the home side
```
{
  "snapshotBlockNumber": 15515659,
  "chainId": 77,
  "RequiredBlockConfirmationChanged": [
    {
      "blockNumber": 9849622,
      "returnValues": {
        "requiredBlockConfirmations": "1"
      }
    }
  ],
  "validatorAddress": "0x0c7A04cc9B1fF1184c5bc7253869727f29593465",
  "RequiredSignaturesChanged": [
    {
      "blockNumber": 9849617,
      "returnValues": {
        "requiredSignatures": "1"
      }
    },
    {
      "blockNumber": 15503549,
      "returnValues": {
        "requiredSignatures": "2"
      }
    }
  ],
  "ValidatorAdded": [
    {
      "blockNumber": 9849617,
      "returnValues": {
        "validator": "0x5d1486951B03A4a8AAdbd9956709041B48003Ea0"
      },
      "event": "ValidatorAdded"
    },
    {
      "blockNumber": 15501966,
      "returnValues": {
        "validator": "0xC5f399f0110E62B9F98fA7e48b66E49216daA598"
      },
      "event": "ValidatorAdded"
    },
    {
      "blockNumber": 15501970,
      "returnValues": {
        "validator": "0x37d95122Cc03CF76a52DEd96f03E51F10c655635"
      },
      "event": "ValidatorAdded"
    }
  ],
  "ValidatorRemoved": []
}
```
It includes all the information related to the bridge contracts that the ALM needs. The only information not stored is about the validator operations transactions, tx receipts, and blocks data. I can try to estimate the growth if we include all the transactions from the validator to sign and execute the messages. However, I don't think storing this information in the snapshot could be useful. Since the snapshot is generated one time when the app is built, users will come later with probably new transactions as they use the bridge which won't be included in the snapshot.

alm/src/services/SnapshotProvider.ts
akolotov: got you, thanks

alm/src/utils/contract.ts
akolotov: Imagine that the bridge has lost the control over the validators contract e.g. the validators contract upgradability admin is not the same account as the bridge upgradability admin AND the private key of the validators contract upgradability admin was lost. So, we will need to stop the bridge and deploy another validator contract with the same parameters as an old one except the validators contract upgradability admin account.

alm/package.json
akolotov: thanks

alm/src/utils/web3.ts
akolotov: OK. I see you point. The contract addresses will be mostly different for another chain so most probably the ALM will not work at all.

alm/src/services/SnapshotProvider.ts
akolotov: Yeah, OK. You are right -- adding this info to the snapshot has no much sense.

alm/src/utils/contract.ts
patitonar: Thanks for the example!
In that case, as you suggested it makes sense to remove the validator contract address.

If the validator contract address is changed, right now there is no way to get the historic values of the validator contracts addresses. So if a user introduces a transaction that was previous to the change to a new validator contract, we won't be able to get the correct parameters of validator list and required signatures at that moment since we won't be able to get the validator contract of that moment and in the new contract, validators could be different.

Does it make sense to add a new event like `ValidatorContractChanged` when the bridge is initialized? it could be emitted also when the validator contract changes in some upgrade method invoked. Then in the future, ALM could use those events.

akolotov: > So if a user introduces a transaction that was previous to the change to a new validator contract, we won't be able to get the correct parameters of validator list and required signatures at that moment since we won't be able to get the validator contract of that moment and in the new contract, validators could be different.

> Does it make sense to add a new event like ValidatorContractChanged when the bridge is initialized? it could be emitted also when the validator contract changes in some upgrade method invoked. Then in the future, ALM could use those events.

OK. Let's introduce such event in the contracts. But we will not handle this situation in ALM so far since it is too rare.

patitonar: Removed the validator address from the snapshot https://github.com/poanetwork/tokenbridge/pull/382/commits/33c06eedd3637fcfb2c421d4e7ef1ae9e4b97d07
And also created an issue for the event in the contracts https://github.com/poanetwork/tokenbridge-contracts/issues/445

PR #389, opened 2020-07-08
Update bw plugin README and version

k1rill-fedoseev:


burner-wallet-plugin/tokenbridge-bw-exchange/README.md
```diff
@@ -19,16 +23,27 @@ yarn add @poanet/tokenbridge-bw-exchange
 ### Usage
 
 ```javascript
-import { Etc, Wetc, EtcGateway, WETCBridge } from '@poanet/tokenbridge-bw-exchange'
+import {
```
akolotov, line 26: @patitonar what was initial logic of this example? Do you think it is better to have several examples with usage? E.g. on example for ETC/WETC, another is to present that exchanges could be combined: MOON/xMOON and Dai/qDAI.

patitonar: The idea of this example is to show how to import the different resources (assets, gateway, pairs) from the package `@poanet/tokenbridge-bw-exchange` and use them in the burner wallet. I think with one example of where should be each of these resources used in the components of the BW is OK. That assets and exchanges could be combined looks to me more like a description of the BW features than our plugin, but if you think having several simpler examples could help the user to take advantage of it we can do it

akolotov: > That assets and exchanges could be combined looks to me more like a description of the BW features than our plugin, but if you think having several simpler examples could help the user to take advantage of it we can do it

Yes, more examples could make the intent of our plugin more understandable. It could increase adoption. @k1rill-fedoseev could you split this example on two described above?

k1rill-fedoseev: Sure, I split example on two smaller ones, I also added some small comments about used assets and gateways.
Please, take a look once again.

PR #394, opened 2020-07-09
Possibility to send a tx through all provided RPC endpoints

### LINKED ISSUE #336: Redundant transactions through the list of configured RPC nodes
Currently the bridge oracle relies on liveness of an RPC node when sends the confirmation or relays collected signatures. In some cases the node could accept the RPC requests but will not propagate the transaction further. It does not report back to the oracle with any kind of error. That's why the bridge do not activate the feature to re-send the transaction through another RPC node (it is allowed to specify several RPC end points in `COMMON_HOME_RPC_URL` and `COMMON_FOREIGN_RPC_URL` configuration parameters).

Implementation of tracking mechanism for every transaction is costly from development efforts point of view whereas the issue which is going to be addressed is rare. Therefore consider an interim solution:
1) introduce a new parameter ORACLE_TX_REDUNDANCY
2) if this parameter is configured as `true`, the oracle will send `eth_sendRawTransaction` call through all healthy RPC nodes configured in `COMMON_*_RPC_URL` parameter.

k1rill-fedoseev:
Closes #336


oracle/src/tx/sendTx.js
```diff
@@ -2,6 +2,8 @@ const Web3Utils = require('web3-utils')
 const fetch = require('node-fetch')
 const rpcUrlsManager = require('../services/getRpcUrlsManager')
 
+const { ORACLE_TX_REDUNDANCY } = process.env
```
akolotov, line 5: Please update the documentation files related to the oracle with this parameter. Please also check that it will be propagated to the target system automatically by the ansible deployment scripts.

oracle/src/services/RpcUrlsManager.js
```diff
@@ -15,19 +19,22 @@ function RpcUrlsManager(homeUrls, foreignUrls) {
   this.foreignUrls = foreignUrls.split(',')
 }
 
-RpcUrlsManager.prototype.tryEach = async function(chain, f) {
+RpcUrlsManager.prototype.tryEach = async function(chain, f, redundant = false) {
```
akolotov, line 22: should it be covered by somehow in the tests?

oracle/src/tx/sendTx.js
k1rill-fedoseev: I have updated corresponding docs.
Regarding Ansible deployment, if I am not mistaken, it should be enough to add `ORACLE_TX_REDUNDANCY` in `deployment/roles/oracle/templates/.env.j2`.

oracle/src/services/RpcUrlsManager.js
k1rill-fedoseev: I added tests for `promiseAny` function, similar to the tests we have for `tryEach` util function. I think this should be enough.

PR #395, opened 2020-07-09
Speed up e2e building process

k1rill-fedoseev:
When running e2e tests (i. e. oracle-e2e, monitor-e2e), it takes a lot of time to build all docker images. I noted that some built images are not used in the specific e2e jobs, so they can be skipped from the building process.


k1rill-fedoseev:
I think we should see the exact numbers and timings on CI, when it will be available again. I didn't run all tests locally, but from my expectation, when building images for a new version of code, this PR could reduce building time by ~50%. 
For example, for oracle-e2e tests, previously we built 4 similar docker images: `oracle`, `monitor`, `ui`, `e2e`. Now, we should build only `e2e` and `oracle`.

k1rill-fedoseev:
So, I have refactored our CI configurations. Here is what I came up with:
Previously, the situation was the following (typical example of a CI workflow - https://app.circleci.com/pipelines/github/poanetwork/tokenbridge/1841/workflows/30be2cae-0a0a-4173-b1df-267593b62542):
- **docker layer caching**: 13 jobs (`13 * 200 = 2600 credits`)
- **compute time**: 412 minutes (`412 * 10 = 4120 credits`)
- **total**: `2600 + 4120 = 6720 credits` per full workflow run on the develop branch

After updates in this PR, the situation seems to be the following (https://app.circleci.com/pipelines/github/k1rill-fedoseev/tokenbridge-ci/22/workflows/e0ca3872-972e-4285-b851-2460423895a5):
-  **docker layer caching**: 0 jobs (`0 * 200 = 0 credits`)
- **compute time**: 238 minutes (`238 * 10 = 2380 credits`)
- **total**: `0 + 2380 = 2380 credits` per full workflow run on the develop branch

In this PR, e2e docker images (`e2e`, `monitor`, `oracle`, `ui`, `molecule_runner`) are built only once in a separate job and then pushed to the docker registry. Later jobs pulls needed built images from the docker registry instead of building them from once again.
I also refactored Dockerfiles in a way that Solidity contracts are now built only once. Built artefacts are then cached and shared between `monitor`, `oracle`, `ui` docker images.

Docker layer caching can be enabled for `build-e2e-images` if needed, it may help to save some minutes for this particular job on its subsequent runs.

Further improvements may consist of restricting some heavy jobs for particular branches. I think it make sense to restrict ultimate tests to `master`/`develop + master` branches only.

akolotov:
@patitonar could take a look at the suggested changes carefully? Do you see any missed parts or things that could be impacted?

patitonar:
This looks really good! can we also add the improvement related to contracts compiles to `alm/Dockerfile`? It will improve the build time of it and also it will be ready for later when we add the alm e2e tests

k1rill-fedoseev:
Yes, sure, we can. Please take a look at the updated `alm/Dockerfile`

PR #397, opened 2020-07-13
Properly handle not found transaction and disable auto refresh

### LINKED ISSUE #392: If a non-existing transactions is provided in ALM the loading is stuck forever
It makes sense to properly handle if no transactions are found and display a message accordingly

### LINKED ISSUE #393: Avoid reloading the ALM page when triggered from other browser components
Right now if Metamask or Nifty wallet triggers a reload of the pages because of a network change, the ALM page also gets reloaded. It makes sense to prevent this default behavior

patitonar:
This PR includes 2 fixes:
- Display the correct message and back button if transaction not found. Closes #392 
- Disable auto refresh feature from injected wallets. Closes #393 

Demo updated: https://alm-demo-e0998.web.app/

### DIFF
--- a/alm/src/components/Form.tsx
+++ b/alm/src/components/Form.tsx
@@ -35,8 +35,13 @@ export const Form = ({ onSubmit }: { onSubmit: ({ chainId, txHash, receipt }: Fo
     onSubmit({ chainId, txHash, receipt })
   }
 
+  const onBack = () => {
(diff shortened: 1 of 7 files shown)


patitonar:
Thanks for the suggestion. I updated the space, similar to the distance from the text to the header. I also updated the demo page
space

PR #407, opened 2020-07-23
Add alm basic e2e tests

### LINKED ISSUE #360: Add e2e tests for ALM
We should add some e2e tests covering different scenarios

patitonar:
Closes #360 

This PR adds some basics e2e tests for ALM to make sure that the app is able to build and run correctly using the docker configuration.

The tests check that the status of the transactions is correctly set in the app for transactions from both directions. To generate some transactions to use, the script first set up the components to start and run the amb oracle e2e tests.


akolotov:
@k1rill-fedoseev please take a look if it complies with the changes you did previously.

patitonar:
@k1rill-fedoseev thanks for the suggestions, I applied the changes. Please take a look

PR #408, opened 2020-07-25
Change the remote RPC url for Ethereum Classic

cryptoasuka:
also https://wetc.app is broken :(


akolotov:
@cryptoasuka please note that after migration of WETC bridge to the AMB protocol, the old application (`https://wetc.app`) is not applicable any more for the token transfer. Please use [a DApp based on the Burner Wallet 2 UI ](https://docs.tokenbridge.net/amb-bridge/create-a-burner-wallet-plugin-for-your-amb-bridge-extension)(e.g. https://wetc-app.herokuapp.com/) instead.

PR #416, opened 2020-08-05
Correction of the message for a transaction without the bridge requests

akolotov:
These changes are needed to get rid of an ambiguity for the case when the transaction was sent to the bridge but did not initiate a bridge message by some reason.

image


akolotov:
> I have fixed small linter errors raised from `npm run lint`, now it LGTM

thanks

PR #419, opened 2020-08-18
fix bug with method handler crashed

sevenzing:
- added MIN_AMB_HEADER_LENGTH to constants.js
- returns constant gas limit for executeAffirmation and submitSignature in case of `method handler crashed`

### DIFF
--- a/oracle/src/events/processAMBAffirmationRequests/estimateGas.js
+++ b/oracle/src/events/processAMBAffirmationRequests/estimateGas.js
@@ -7,7 +7,8 @@ const { parseAMBHeader } = require('../../utils/message')
 const { strip0x } = require('../../../../commons')
 const {
   AMB_AFFIRMATION_REQUEST_EXTRA_GAS_ESTIMATOR: estimateExtraGas,
-  MIN_AMB_HEADER_LENGTH
+  MIN_AMB_HEADER_LENGTH,
+  FALLBACK_GAS_ESTIMATE
 } = require('../../utils/constants')
 
 async function estimateGas({ web3, homeBridge, validatorContract, message, address }) {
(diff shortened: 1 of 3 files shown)


k1rill-fedoseev (commented):
Could you apply a similar patch for code parts related to the other event handlers? Seems that the same issue can occur during a call to the `eth_estimateGas` in these places as well:
https://github.com/poanetwork/tokenbridge/blob/c2f6b5e8ba65eae06f876f26960e790c93ce986c/oracle/src/events/processAffirmationRequests/estimateGas.js#L9
https://github.com/poanetwork/tokenbridge/blob/c2f6b5e8ba65eae06f876f26960e790c93ce986c/oracle/src/events/processAMBCollectedSignatures/estimateGas.js#L25
https://github.com/poanetwork/tokenbridge/blob/c2f6b5e8ba65eae06f876f26960e790c93ce986c/oracle/src/events/processCollectedSignatures/estimateGas.js#L23

akolotov:
> Could you apply a similar patch for code parts related to the other event handlers? Seems that the same issue can occur during a call to the `eth_estimateGas` in these places as well:

Let's discuss... From my perspective these changes are not needed due to several reasons:
1. The changes are being made under this PR are for demonstration purposes of the AMB bridge between xDai and AVA C-chain. The Home chain is the AVA C-chain. The issue will not happen for the other bridge modes. Also the issue will not happen on the foreign side since we don't plan to use the AVA C-chain as the Foreign chain.
2. The PR is a workaround for the issue needs to be fixed on the side of the particular client. As soon as the fix is appeared on the client side these changes are not necessary anymore. So, more changes we introduce now, more changes must reverted then.
3. This PR is for review only, most probably the changes will live in Lev's fork of the repo.

PR #425, opened 2020-09-04
Possibility to resend old pending transactions

k1rill-fedoseev:


oracle/src/sender.js
```diff
@@ -114,11 +122,37 @@ async function main({ msg, ackMsg, nackMsg, channel, scheduleForRetry }) {
       }
 
       try {
-        logger.info(`Sending transaction with nonce ${nonce}`)
+        let txNonce
+        if (isResend) {
+          const tx = await web3Instance.eth.getTransaction(job.txHash)
```
akolotov, line 127: Imagine, that we have 10 tx in the sent queue and 1 new transaction to be sent. Does it mean, the the sender worker will spent time first to get the transaction info 10 times and only after that send the new transaction. What if we have 100 txs in the sent queue? Do you think it is better to introduce a separate worker that will work with the sent queue as so the sender worker will not be blocked by monitoring the old txes?

k1rill-fedoseev: I get your point, I agree that it is indeed an issue. However, running separate workers for home and foreign sides looks for me as a huge overhead.
Seems like we can solve this problem by using priority queues. If any regular job for sending a transaction will have a higher priority than a similar job for checking sent transactions statuses, sender worker will always choose to send a new tx first instead of processing an already sent transaction. 
What do you think?

k1rill-fedoseev: **Important Note**:
It is not possible to change the parameters of the existing queues in RabbitMQ, without deleting it and all stored messages. So, instead, I create a new queues `home-prioritized` and `foreign-prioritized` (instead of `home` and `foreign`) and then redirect any remaining messages from old queues to the new ones during the sender's startup at: https://github.com/poanetwork/tokenbridge/blob/deaa10b3e5ebc3e3a7a231198047ed6cae5f356b/oracle/src/services/amqpClient.js#L75

Such approach should allow to upgrade existing oracles without losing existing messages.

akolotov (commented):
Are any upgrade actions required if a worker with these changes starts functioning on the system where the bridge oracle was already deployed and run?

k1rill-fedoseev:
> Are any upgrade actions required if a worker with these changes starts functioning on the system where the bridge oracle was already deployed and run?

No, I think everything should be updated smoothly, without additional actions required.

akolotov (approved):

PR #442, opened 2020-09-24
Possibility to allow/block specific addresses in erc-to-native mode

k1rill-fedoseev:
This PR allows validator to create a special files, `allowance_list.txt` and `block_list.txt`, which will filter out some of the CollectedSignatures requests.
* If `allowance_list.txt` is not empty, then only requests with either sender or recipient being in the list will be processed.
* If `block_list.txt` is not empty, then only requests with neither sender nor recipient being in this list will be processed.
* In all other cases, every request is processed as it is.


oracle/src/events/processCollectedSignatures/index.js
```diff
@@ -46,6 +47,34 @@ function processCollectedSignaturesBuilder(config) {
         logger.info(`Processing CollectedSignatures ${colSignature.transactionHash}`)
         const message = await homeBridge.methods.message(messageHash).call()
 
+        if (config.accessLists) {
+          const parsedMessage = parseMessage(message)
+          const recipient = parsedMessage.recipient.toLowerCase()
+          const originalTxHash = parsedMessage.txHash
+          const allowanceList = await readAccessListFile(config.accessLists.allowanceList)
+          const blockList = await readAccessListFile(config.accessLists.blockList)
```
akolotov, line 55: Does it make sense to read the file once when the worker is started? I understand that it is good to have the file re-read automatically but if it is not changed to often so we will get it opened again and again every for transaction. This is especially not efficient if there is no file at all.

oracle/src/events/processCollectedSignatures/index.js
```diff
@@ -46,6 +47,34 @@ function processCollectedSignaturesBuilder(config) {
...
+
+          if (allowanceList.length > 0) {
+            const sender = (await web3Home.eth.getTransaction(originalTxHash)).from.toLowerCase()
```
akolotov, line 58: I would suggest an optimization here: to check the sender only If the recipient is not found in the list. In most cases it will be enough since only few users understand how to use the alternative recipient feature.

oracle/src/events/processCollectedSignatures/index.js
```diff
@@ -46,6 +47,34 @@ function processCollectedSignaturesBuilder(config) {
...
+                'Validator skips a transaction. Neither sender nor recipient addresses are in the allowance list.',
+                { sender, recipient }
+              )
+              return
+            }
+          } else if (blockList.length > 0) {
+            const sender = (await web3Home.eth.getTransaction(originalTxHash)).from.toLowerCase()
+            if (blockList.indexOf(recipient) > -1 || blockList.indexOf(sender) > -1) {
+              logger.info(
+                'Validator skips a transaction. Neither sender nor recipient addresses are in the block list.',
```
akolotov, line 70: The message is copied from code that handles the allowance list so it does not reflect the nature of the block list. I believe it should be "Validator skips a transaction. Either sender or recipient addresses are in the block list."

oracle/src/events/processCollectedSignatures/index.js
```diff
@@ -46,6 +47,34 @@ function processCollectedSignaturesBuilder(config) {
...
+          if (allowanceList.length > 0) {
+            const sender = (await web3Home.eth.getTransaction(originalTxHash)).from.toLowerCase()
+            if (allowanceList.indexOf(recipient) === -1 && allowanceList.indexOf(sender) === -1) {
+              logger.info(
+                'Validator skips a transaction. Neither sender nor recipient addresses are in the allowance list.',
+                { sender, recipient }
+              )
+              return
+            }
+          } else if (blockList.length > 0) {
```
akolotov, line 66: Let's change the logic: check the block list first and if it is empty then the allowance list should be  handled. The reason is that the block list is going to be used more often.

oracle/config/collected-signatures-watcher.config.js
```diff
@@ -2,11 +2,20 @@ const baseConfig = require('./base.config')
...
   ...baseConfig.homeConfig,
   event: 'CollectedSignatures',
   queue: 'foreign-prioritized',
   name: `watcher-${id}`,
   id
 }
+
+if (baseConfig.id === 'erc-native') {
+  config.accessLists = {
+    blockList: '/mono/oracle/access-lists/block_list.txt',
```
akolotov, line 16: Does it make sense to specify the path in the configuration parameters provided in `.env`? E.g. two parameters could be introduced `ORACLE_HOME_TO_FOREIGN_ALLOWANCE_LIST` and `ORACLE_HOME_TO_FOREIGN_BLOCK_LIST`. If the parameter is not configured, don't try to read the file at all.

akolotov (changes requested):

oracle/src/events/processCollectedSignatures/index.js
k1rill-fedoseev: Initially, I thought that it is required to support file updates in real time. But if it is not needed, then we can save file contents in memory.

oracle/src/events/processCollectedSignatures/index.js
k1rill-fedoseev: Sure, makes sense.

oracle/src/events/processCollectedSignatures/index.js
k1rill-fedoseev: Sure, addressed in the last commit.

oracle/config/collected-signatures-watcher.config.js
k1rill-fedoseev: I think it does not make too much sense. In any case, we are running our services in docker containers, so these file names are only for the internal usage within a container.

akolotov: But if you run it locally by yarn and without the docker, you need to create directory `/mono/oracle/access-lists` on your system.

oracle/src/events/processCollectedSignatures/index.js
akolotov: I would prefer to cache the content. The idea is that we will change the file not often (ideally allowance list must not be used at all) so it is OK to restart the service instead of having extra read cycles during the the worker operations.

oracle/src/events/processCollectedSignatures/index.js
```diff
@@ -46,6 +47,36 @@ function processCollectedSignaturesBuilder(config) {
...
 
+        if (config.accessLists) {
+          const parsedMessage = parseMessage(message)
+          const recipient = parsedMessage.recipient.toLowerCase()
+          const originalTxHash = parsedMessage.txHash
+          const allowanceList = await readAccessListFile(config.accessLists.allowanceList)
+          const blockList = await readAccessListFile(config.accessLists.blockList)
+
+          if (blockList.length > 0) {
+            const sender = (await web3Home.eth.getTransaction(originalTxHash)).from.toLowerCase()
```
akolotov, line 58: It makes sense to implement here the logic similar to the allowance list: check existence of the recipient in the list and only after that get the transaction details.

akolotov (changes requested):

oracle/config/collected-signatures-watcher.config.js
k1rill-fedoseev: Ok, introduced two environment variables for that.

oracle/src/events/processCollectedSignatures/index.js
k1rill-fedoseev: Addressed in the 05144cd.

oracle/src/events/processCollectedSignatures/index.js
k1rill-fedoseev: Added in-memory caching

oracle/src/events/processCollectedSignatures/index.js
```diff
@@ -46,6 +49,37 @@ function processCollectedSignaturesBuilder(config) {
...
 
+        if (ORACLE_HOME_TO_FOREIGN_ALLOWANCE_LIST || ORACLE_HOME_TO_FOREIGN_BLOCK_LIST) {
+          const parsedMessage = parseMessage(message)
+          const recipient = parsedMessage.recipient.toLowerCase()
+          const originalTxHash = parsedMessage.txHash
+
+          if (ORACLE_HOME_TO_FOREIGN_ALLOWANCE_LIST) {
+            const allowanceList = await readAccessListFile(ORACLE_HOME_TO_FOREIGN_ALLOWANCE_LIST)
+            if (allowanceList.indexOf(recipient) === -1) {
+              const sender = (await web3Home.eth.getTransaction(originalTxHash)).from.toLowerCase()
```
akolotov, line 60: To get sender for large amount of transactions to check the address allowance and for almost every transaction to check the blocked address seems for me as a quite costly operation. Can we have additional option to enable this check, e.g. `ORACLE_HOME_TO_FOREIGN_CHECK_SENDER`? This parameter could be omitted in the configuration file so it is considered as `false`.

akolotov, line 60: Can we add a debug message that the transaction is requested to get withdrawal request sender? The same of the block list...

oracle/src/events/processCollectedSignatures/index.js
```diff
@@ -55,27 +59,35 @@ function processCollectedSignaturesBuilder(config) {
...
+                const sender = (await web3Home.eth.getTransaction(originalTxHash)).from.toLowerCase()
+                if (allowanceList.indexOf(sender) === -1) {
+                  logger.info(
+                    { sender, recipient },
+                    'Validator skips a transaction. Neither sender nor recipient addresses are in the allowance list.'
+                  )
+                  return
+                }
+              } else {
                 return
```
akolotov, line 75:
```suggestion
                logger.info(
	                { sender, recipient },
                  'Validator skips a transaction. Recipient address is not in the allowance list.'
                )
                return
```

oracle/src/utils/utils.js
```diff
@@ -107,7 +107,8 @@ const invert = p => new Promise((res, rej) => p.then(rej, res))
 const promiseAny = ps => invert(Promise.all(ps.map(invert)))
 
 const readAccessLists = {}
-async function readAccessListFile(fileName) {
+async function readAccessListFile(fileName, logger) {
+  logger.debug({ fileName }, 'Access list file read requested')
```
akolotov, line 111: I would either move this line to be within the "if" statement or, if the filename was not configured, report about this.

akolotov (changes requested):

akolotov (approved):

PR #443, opened 2020-09-25
Reduce the number of JSON-RPC calls in UI

k1rill-fedoseev:


ui/src/stores/HomeStore.js
```diff
@@ -403,12 +406,14 @@ class HomeStore {
 
   @action
   async filterByTxHashInReturnValues(transactionHash) {
-    const events = await this.getEvents(1, 'latest')
+    const { blockNumber } = await this.homeWeb3.eth.getTransaction(transactionHash)
```
akolotov, line 409: What if the transaction is not mined yet?

akolotov (commented):
Did you check this changes on the real deposit and withdrawal? Does the UI check the transaction status properly?

ui/src/stores/HomeStore.js
k1rill-fedoseev: Indeed, there was a mistake. I have fixed it, and checked the ui in local testnet, everything seems be working fine.

ui/src/stores/HomeStore.js
```diff
@@ -406,14 +406,17 @@ class HomeStore {
 
   @action
   async filterByTxHashInReturnValues(transactionHash) {
-    const { blockNumber } = await this.homeWeb3.eth.getTransaction(transactionHash)
-    const events = await this.getEvents(blockNumber, blockNumber)
+    const events = await this.getEvents(1, 'latest')
```
akolotov: Does it means that the startBlock is `1`? Do you think it makes sense to choose more relevant block? E.g. a block that was mined a day ago from the current moment or something like this?

k1rill-fedoseev: But what if a user want to access the events for the transaction happened more than 1 day ago?

akolotov: Is this for scenario when a user accesses a tx in the event tab? Will this tab display not mined transactions? If it is not from this tab, how the will the user access this transaction?

ui/src/stores/ForeignStore.js
```diff
@@ -168,9 +169,9 @@ class ForeignStore {
     setInterval(() => {
       this.getBlockNumber()
       this.getEvents()
```
akolotov, line 171: For which scenario are data/actions produced getEvents() used? Is it worth to reduce amount of requests to frame the getting events only for particular cases?

ui/src/stores/HomeStore.js
k1rill-fedoseev: This function is used for searching home side events which include transactionHash on the foreign side in their data (such as `AffirmationCompleted`, `SignedForAffirmation`)

akolotov (approved):

PR #444, opened 2020-09-28
Update the condition for checking if a tx has stucked

k1rill-fedoseev:
It was found that Infura providers return `null` for pending transactions instead of the tx object with `"blockNumber": null`. This PR updates the behaviour for stucked transactions and makes the checking logic more universal.

### DIFF
--- a/oracle/src/sender.js
+++ b/oracle/src/sender.js
@@ -127,7 +127,7 @@ async function main({ msg, ackMsg, nackMsg, channel, scheduleForRetry, scheduleT
           const tx = await web3Instance.eth.getTransaction(job.txHash)
 
           if (tx && tx.blockNumber !== null) {
-            logger.info(`Transaction ${job.txHash} was successfully mined`)
+            logger.debug(`Transaction ${job.txHash} was successfully mined`)
             return
           }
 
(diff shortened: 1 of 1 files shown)


oracle/src/sender.js
```diff
@@ -123,53 +123,44 @@ async function main({ msg, ackMsg, nackMsg, channel, scheduleForRetry, scheduleT
...
         if (isResend) {
           const tx = await web3Instance.eth.getTransaction(job.txHash)
 
-          if (tx === null) {
-            logger.info(`Transaction ${job.txHash} was not found, dropping it`)
-            return
-          }
-          if (tx.blockNumber !== null) {
+          if (tx && tx.blockNumber !== null) {
             logger.info(`Transaction ${job.txHash} was successfully mined`)
```
akolotov, line 130:
```suggestion
            logger.debug(`Transaction ${job.txHash} was successfully mined`)
```

akolotov: Since we optimistically assume that most transactions will be mined, so, this will overload the logs.

akolotov (changes requested):

akolotov (approved):

PR #459, opened 2020-10-01
Fix insufficient funds error handler

k1rill-fedoseev:


akolotov (commented):
As per internal discussion it is necessary to revisit the solution in order to cover all possible cases

akolotov (changes requested):

PR #467, opened 2020-10-06
Fix isBridgeContract handler in burner wallet plugin

### LINKED ISSUE #445: Instable connection to RPC causes incorrect operations with tokens in BW plugin
It has been discovered that the following code returns `false` if a RPC endpoint used by the burner wallet core return errors (e.g. quota limit for the infura endpoint):

https://github.com/poanetwork/tokenbridge/blob/48dd53622c5b13cef5d9ae923cc148eb77b61d0a/burner-wallet-plugin/tokenbridge-bw-exchange/src/burner-wallet/assets/BridgeableERC20Asset.ts#L33-L43

It causes that the token is sent to the bridge/mediator contract by the `transfer` method instead of the pair of methods `approve` + `relayTokens`

k1rill-fedoseev:
Closes #445

### DIFF
--- a/burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/utils.ts
+++ b/burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/utils.ts
@@ -51,6 +51,9 @@ export const isBridgeContract = async (contract: Contract, allowedModes?: string
     }
     return allowedModes.includes(mode)
   } catch (e) {
-    return false
+    if (e.message.includes("Returned values aren't valid")) {
+      return false
+    }
+    throw e
   }
 }


burner-wallet-plugin/tokenbridge-bw-exchange/src/utils/utils.ts
```diff
@@ -51,6 +51,9 @@ export const isBridgeContract = async (contract: Contract, allowedModes?: string
     }
     return allowedModes.includes(mode)
   } catch (e) {
-    return false
+    if (e.message.includes("Returned values aren't valid")) {
```
akolotov, line 54: What is your advise to verify this functionality with the Infura or xDai rpc endpoints?

k1rill-fedoseev: Well, I would suggest to add an additional `console.log` statement here, which will print out the `e.message`, so that we can see what kind of errors get in different situations.
In my local experiments, the error starting from `Returned values aren't valid` happened only when the callee was either an EOA or some non-bridge contract. Also, seems that this error is raised from the `web3js` library, so the error message should not depend on the particular JSON-RPC implementation.

akolotov (approved):

akolotov:
@k1rill-fedoseev should we update the corresponding NPM package and reference to it here: https://github.com/poanetwork/tokenbridge-burner-wallet-plugin/blob/71e816b40104d2a0b1bccac4afbb2023f686efe1/yarn.lock#L2361?

PR #475, opened 2020-10-21
Reduce time complexity of monitor

k1rill-fedoseev:
This PR reduces the time complexity for events processing from `O(n^2)` to `O(n)`. Unneeded `keccak256` operations were replaced by case insensitive comparisons as well for the sake of performance .


akolotov (approved):
Tested on our set of bridges. The overall duration was reduced by ~80 seconds. Thanks!

PR #477, opened 2020-10-22
Add allowance/block lists support to monitor

### LINKED ISSUE #454: Adapt monitor to take care of transactions handled by oracle's AL/BL
After integration #442 it could be a situation when the withdrawals are blocked by the oracle and monitor starts reporting about imbalance. This issue is being created to find out a way how to handle this properly on the monitor side.

k1rill-fedoseev:
Closes #454


monitor/utils/message.js
```diff
@@ -64,9 +67,22 @@ const eventWithoutReference = otherSideEvents => e =>
   otherSideEvents.filter(a => a.referenceTx === e.referenceTx && a.recipient === e.recipient && a.value === e.value)
     .length === 0
 
+const unclaimedHomeToForeignRequests = () => {
+  if (MONITOR_HOME_TO_FOREIGN_ALLOWANCE_LIST) {
+    const allowanceList = readAccessListFile(MONITOR_HOME_TO_FOREIGN_ALLOWANCE_LIST)
+    return e => !allowanceList.includes(e.recipient.toLowerCase())
```
akolotov, line 73: The similar functionality in the oracle also depends on the sender -- whether the sender of transaction is in AL/BL. I understand that it could impact the duration how the monitor handles the bridge operations. Could we estimate how duration will be increased? In order to speed up the process does it makes sense to cache the correspondence between a sender and a transaction as so monitor workers do not need to get the transaction data again and again?

monitor/getShortEventStats.js
```diff
@@ -24,9 +29,19 @@ async function main(bridgeMode) {
       }
     }
   } else {
+    const unclaimedStats = {}
+    if (MONITOR_HOME_TO_FOREIGN_ALLOWANCE_LIST || MONITOR_HOME_TO_FOREIGN_BLOCK_LIST) {
+      const unclaimedPool = homeToForeignRequests
+        .filter(eventWithoutReference(homeToForeignConfirmations))
+        .filter(unclaimedHomeToForeignRequests())
+
+      unclaimedStats.unclaimed = unclaimedPool.length
```
akolotov, line 38: consider to decrease `depositsDiff` by `unclaimedStats.unclaimed` as so we can see in `depositsDiff` only the number of relay operations which the oracles must handle but did not so.

monitor/getShortEventStats.js
```diff
@@ -24,9 +29,19 @@ async function main(bridgeMode) {
...
+      unclaimedStats.unclaimedDiff = Web3Utils.fromWei(BN.sum(...unclaimedPool.map(e => e.value)).toFixed())
```
akolotov, line 39: let's call it `unclaimedBalance` and reduce `balanceDiff` by this value. So, we will see: "ok, due to bug in oracle we have such balance diff where as we have unclaimedBalance for the operations that were not finalized by users"

monitor/utils/message.js
k1rill-fedoseev: Well, if the sender of a transaction will be cached, so it is required to fetch it only once, than nothing bad should happen. Assuming that we have 1000 tx per day, each of them will be marked as unclaimed, each request takes ~300ms to execute, in total we have 300 seconds per day, or 13 seconds per hour.
I don't see a big problem here, so everything should be fine.

monitor/checkWorker.js
```diff
@@ -63,6 +65,7 @@ async function checkWorker() {
     vBalances.ok = vBalances.homeOk && vBalances.foreignOk
     vBalances.health = true
     writeFile(`/responses/${MONITOR_BRIDGE_NAME}/validators.json`, vBalances)
+    saveCache()
```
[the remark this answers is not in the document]
k1rill-fedoseev: I don't see the reason why it will be better. I think it makes sense to save the caches on disk at the very end of the script, so that we don't lose any changes in case transaction senders will be fetched in other functions as well. Even though now it is not the case, it can be in the future.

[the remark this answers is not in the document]
akolotov: In general you are right. But in this particular case we request the state of the validators contract in the line 41
```
const vBalances = await validators(bridgeMode)
```
so, if the script fails here due to some exception the cache will not be stored.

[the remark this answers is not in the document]
k1rill-fedoseev: Ok, that makes some sense to me, let's do it this way.

akolotov (approved):

PR #482, opened 2020-10-29
Cache fetched events in monitor

k1rill-fedoseev:


monitor/utils/web3Cache.js
```diff
@@ -16,14 +19,98 @@ async function getHomeTxSender(txHash) {
...
 
+async function getPastEvents(contract, options) {
+  if (MONITOR_CACHE_EVENTS !== 'true') {
+    return commonGetPastEvents(contract, options)
+  }
+
+  const contractAddr = contract.options.address
+  const eventABI = contract.options.jsonInterface.find(e => e.type === 'event' && e.name === options.event)
+  const eventSignature = `${eventABI.name}(${eventABI.inputs.map(i => i.type).join(',')})`
+  const cacheFile = `./cache/${MONITOR_BRIDGE_NAME}/${options.chain}/${contractAddr}/${eventSignature}.json`
```
akolotov, line 30: 1) do you think it is really necessary to have the directory `${MONITOR_BRIDGE_NAME}` within the `cache` directory? As you can see the launcher of monitor scripts `getBridgeStats.sh` already mounts the `cache` directory with respect to the bridge name:
https://github.com/poanetwork/tokenbridge/blob/0228fc7d5f5f8d3ca6d60c5d23d5f76cacc3cd61/monitor/scripts/getBridgeStats.sh#L43

2. Does it make sense to keep the chain id instead of symbolic names 'foreign' and 'home' in `${options.chain}`?

k1rill-fedoseev: 1. I have changed the bash script instead, so that the cache should now work both inside Docker and not.
2. Making such change will require to do 2 extra RPC calls each time worker is running. While it can be fixed by some kind of chain id caching, I think it is simpler to keep a pair of `(bridge name, bridge side)` as a chain identifier.

monitor/getBalances.js
```diff
@@ -18,21 +18,50 @@ const {
...
+
+  // Foreign balance should represent all UserRequestForAffirmation events up to block `N - requiredBlockConfirmation()`
+  // and all RelayedMessage events up to block `N`.
+  // This constant tells the difference between bridge balance at block `N - requiredBlockConfirmation() + 1`
+  // and the actual value monitor is interested in.
+  const lateForeignConfirmationsTotalValue = BN.sum(
+    0,
+    ...homeToForeignConfirmations.filter(e => e.blockNumber > foreignDelayedBlockNumber).map(e => e.value)
+  )
+  // Foreign balance should represent all UserRequestForSignature events up to block `M - requiredBlockConfirmation()`
```
akolotov, line 40:
```suggestion
  // Home balance should represent all UserRequestForSignature events up to block `M - requiredBlockConfirmation()`
```

akolotov (changes requested):

akolotov (approved):

PR #483, opened 2020-10-30
Support of AMB manual lane in the oracle

### LINKED ISSUE #479: Manual lane and oracles driven lane introductions for AMB oracles
The AMB oracle worker that is responsible for handling collected signatures event raised by the AMB contracts must looks for the files that is called "allowance list" and "block list" as so:

  * [ ] if the allowance list is configured and both the message recipient and the message sender are not in this list, the oracles signatures of this message will not be delivered to the Foreign chain
  * [ ] if the block list is configured and either the message recipient or the message sender is in this list, the oracles signatures of this message will not be delivered to the Foreign chain

### LINKED ISSUE #480: Extension controlled manual lane support by the AMB oracles
As soon as https://github.com/poanetwork/tokenbridge/issues/479 and https://github.com/poanetwork/tokenbridge-contracts/issues/542 are implemented, the corresponding changes must be done on the AMB oracle side.

The AMB oracle worker that is responsible for handling the collected signatures event must check if the relaying message contains the 8th bit of the `dataType` field set and don't transfer the oracles signatures to the Foreign chain even if either the message recipient or the message sender is in the allowance list. The check of `dataType` must be performed only if there is at least one address configured in the allowance list.

k1rill-fedoseev:
Closes #479 
Closes #480


oracle/src/events/processAMBCollectedSignatures/index.js
```diff
@@ -5,14 +5,20 @@ const bridgeValidatorsABI = require('../../../../contracts/build/contracts/Bridg
...
 const { AlreadyProcessedError, IncompatibleContractError, InvalidValidatorError } = require('../../utils/errors')
 const { MAX_CONCURRENT_EVENTS, EXTRA_GAS_ABSOLUTE } = require('../../utils/constants')
 
 const limit = promiseLimit(MAX_CONCURRENT_EVENTS)
 
-const { ORACLE_ALWAYS_RELAY_SIGNATURES } = process.env
+const {
+  ORACLE_HOME_TO_FOREIGN_ALLOWANCE_LIST,
+  ORACLE_HOME_TO_FOREIGN_BLOCK_LIST,
+  ORACLE_HOME_TO_FOREIGN_CHECK_SENDER,
```
akolotov, line 19: this parameter was introduced for the vanilla bridge to provide an ability to turn on/off a costly check of the sender. In the AMB case, this check is chip since the sender is stored in the AMB message. Consider to remove this logic from AMB code.

oracle/src/events/processAMBCollectedSignatures/index.js
```diff
@@ -50,6 +56,48 @@ function processCollectedSignaturesBuilder(config) {
...
+              return
+            }
+            if (ORACLE_HOME_TO_FOREIGN_CHECK_SENDER === 'true' && blockList.includes(sender)) {
+              logger.info({ sender }, 'Validator skips a message. Sender address is in the block list.')
+              return
+            }
+          }
+        }
+
+        if (parsedMessage.dataType & 128) {
```
akolotov, line 94: Does it make sense to wrap it by a function to have a meaningful name like 'isLaneManagedByExtension'? I would also recommend to add a comment to not forget why this bit is being checked.

akolotov (changes requested):

oracle/src/events/processAMBCollectedSignatures/index.js
k1rill-fedoseev: Done in 2e53a71

oracle/src/events/processAMBCollectedSignatures/index.js
k1rill-fedoseev: Done in 5437d63

akolotov (approved):

PR #484, opened 2020-11-03
Support AMB manual lane and allowance/block lists in the monitor

### LINKED ISSUE #481: Reflect messages forwarded to manual lane in monitor
After introduction of #479 and #480 it will be necessary to change the monitor as so it differentiates messages in the manual lane (`fromHomeToForeignPBUDiff`) and the oracle driven lane (`fromHomeToForeignDiff`). `PBU` means "gas fees Paid By Users"

k1rill-fedoseev:
Closes #481


monitor/utils/message.js
```diff
@@ -66,10 +66,32 @@ const unclaimedHomeToForeignRequests = () => {
...
+      return (
+        decodedDataType.manualLane ||
+        (!allowanceList.includes(sender.toLowerCase()) && !allowanceList.includes(executor))
+      )
+    }
+  } else if (MONITOR_HOME_TO_FOREIGN_BLOCK_LIST) {
+    const blockList = readAccessListFile(MONITOR_HOME_TO_FOREIGN_BLOCK_LIST)
+    return e => {
+      const { sender, executor, decodedDataType } = normalizeAMBMessage(e)
+      return decodedDataType.manualLane || blockList.includes(sender.toLowerCase()) || blockList.includes(executor)
```
akolotov, line 83: The monitor needs to consider a message as forwarded to the manual lane regardless to `decodedDataType.manualLane` value: if either an account sender or a contract receiver are in the block list, the oracles don't trust it -- so, they don't take into consideration the value of the `decodedDataType.manualLane` bit.

monitor/utils/message.js
```diff
@@ -66,10 +66,32 @@ const unclaimedHomeToForeignRequests = () => {
   }
 }
 
+const manuallyProcessedAMBHomeToForeignRequests = () => {
+  if (MONITOR_HOME_TO_FOREIGN_ALLOWANCE_LIST) {
+    const allowanceList = readAccessListFile(MONITOR_HOME_TO_FOREIGN_ALLOWANCE_LIST)
+    return e => {
+      const { sender, executor, decodedDataType } = normalizeAMBMessage(e)
+      return (
+        decodedDataType.manualLane ||
```
akolotov, line 75: Is this more logical to check this flag only if `(!allowanceList.includes(sender.toLowerCase()) && !allowanceList.includes(executor))` is false?

monitor/utils/message.js
k1rill-fedoseev: Yes, that is the desired behaviour. That's why we have two OR conditions here. If either manualLane bit is set, or some participating address is blocked, the message is considered as forwarded to the manual lane.

monitor/utils/message.js
k1rill-fedoseev: I don't see the difference between these options. Why is it more logical?

monitor/utils/message.js
akolotov: Ok. I see now. Thanks.

monitor/utils/message.js
akolotov: Because it reflects the logic which we put behind these code changes: for the senders/receivers included in the allowance list we check if the extension set the `manualLane` flag.

Even from performance point of view we can say that there could be bunch extensions included in the allowance list and only few of them in relatively small number of transactions will set the `manualLane` flag. 

Here is an illustration:
For example, we have 10 extensions deployed for AMB, we include in the allowance list only two of them. So, it means that we don't need to check the `manualLane` flag for 8 extensions at all but the suggested code will do this.

k1rill-fedoseev: Ok, now it makes some sense for me. Thanks.

akolotov (approved):

PR #493, opened 2020-11-13
Detect all AMB mediators in monitor

### LINKED ISSUE #295: Detect all pair of mediators for AMB by monitor
The current stats for AMB usage in the tokenbridge monitor does not allow to discover mediator pairs which use the bridge. This information could be useful in order to understand popularity of the bridge.

This functionality should handle properly a case when a user uses the AMB bridge directly without the mediators involvement (e.g. https://etherscan.io/tx/0x84c9e26199e26137ff945b7b35fcf727736183c165fa91388c2920f695907725).

k1rill-fedoseev:
Closes #295

### DIFF
--- a/monitor/utils/web3Cache.js
+++ b/monitor/utils/web3Cache.js
@@ -1,6 +1,6 @@
 const logger = require('../logger')('web3Cache')
 const { readCacheFile, writeCacheFile } = require('./file')
-const { web3Home } = require('./web3')
+const { web3Home, web3Foreign } = require('./web3')
 const { getPastEvents: commonGetPastEvents } = require('../../commons')
 
 const { MONITOR_BRIDGE_NAME, MONITOR_CACHE_EVENTS } = process.env
@@ -9,6 +9,10 @@ let isDirty = false
 
 const homeTxSendersCacheFile = `./cache/${MONITOR_BRIDGE_NAME}/home/txSenders.json`
 const cachedHomeTxSenders = readCacheFile(homeTxSendersCacheFile) || {}
+const homeIsContractCacheFile = `./cache/${MONITOR_BRIDGE_NAME}/home/isContract.json`
+const cachedHomeIsContract = readCacheFile(homeIsContractCacheFile) || {}
+const foreignIsContractCacheFile = `./cache/${MONITOR_BRIDGE_NAME}/foreign/isContract.json`
+const cachedForeignIsContract = readCacheFile(foreignIsContractCacheFile) || {}
 
 async function getHomeTxSender(txHash) {
   if (!cachedHomeTxSenders[txHash]) {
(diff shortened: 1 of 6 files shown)


monitor/utils/web3Cache.js
```diff
@@ -19,6 +23,24 @@ async function getHomeTxSender(txHash) {
   return cachedHomeTxSenders[txHash]
 }
 
+async function isHomeContract(address) {
+  if (typeof cachedHomeIsContract[address] !== 'boolean') {
+    logger.debug(`Fetching home contract code size for tx ${address}`)
+    cachedHomeIsContract[address] = (await web3Home.eth.getCode(address)).length > 2
```
akolotov, line 29: Will it be still OK when a contract was destroyed but our cache thinks that is not?

k1rill-fedoseev: I think it should be totally fine in that case, since any Ethereum address is either an EOA, or a contract (might be not deployed yet, deployed, or destroyed). If our cache remembers that some address belonged to a contract at some point back in time, then there is almost no way that someone has a private key from this address.

k1rill-fedoseev: However, things are not that smooth in case of `executor` address, since any address can be specified as the `executor`. Maybe it makes sense to check the code size only for the sender entity?

akolotov: Let's think more...

In fact, we are interesting only in AMB extensions. I would say that the extension is the pair of mediators which are contracts. If the same pair of mediators contracts addresses are used in more than one transaction we can call it a permanent extension. If one mediator sends messages to different recipients but they are contracts, it is a floating extension. We can check that the recipient is a contract by getting the result of the logical operation (the status of the message execution || the contract's code exists). 

The situation when the sender is EOA and it calls the contract on another side of the bridge could be called an rc-extension (remotely controlled). It can be identified by checking the sender of the originating transaction - it must be the same as the AMB message sender. The case when both sender and recipient could be named a EOAs messages exchange but we have no way to identify it precisely. So, we can just mark all cases when the status of the message execution is failed and the recipient has no code, as 'unknown'.

To sum up... consider to have four sections in `mediators.json`: PermanentExtensions, FloatingExtensions, RCExtensions, Unknown. Use approaches described above to identify the type of the sender and the recipient.

k1rill-fedoseev: Ok, It is clear now how to distinguish between contracts and EOA on both sides.

So we can introduce the following structuring rules for the interactions:
* Home contract A sent messages to Foreign contract B && Foreign contract B sent messages to contract A => Permanent extension
* Home/Foreign contract A sent messages to one or more contracts on the opposite side, without getting any messages back => Floating extension
* Home/Foreign EOA sent messages to the contract on the opposite side => RemotelyControlled extensions
* Messages where destination in an EOA => Unknown

To be more concrete, let's take an example. `HC*` means Home Contract, `FC*` means Foreign Contract, `HU*` means Home User EOA, `FU*` means Foreign User EOA.
Assume the following AMB messages were executed:
* `HC1 => FC1`, `HC1 => FC2`, `HC1 => FC3`, `HC1 => FU1`
* `FC1 => HC1`
* `FC2 => HC2`
* `HU1 => FC1`, `HU1 => FC2`, `HU1 => FU1`
* `FU1 => HC1`, `FU1 => HU1`

The same messages in a form of a graph:
Screenshot 2020-11-20 at 12 24 43

In this case, the following relationships may exist:
* Permanent extension `HC1 <=> FC1`
* Floating extension `HC1 => FC2, FC3`
* Floating extension `FC2 => HC2`
* RC extension `HU1 => FC1`
* RC extension `HU1 => FC2`
* RC extension `FU1 => HC1`
* Unknown extension `HC1 => FU1`
* Unknown extension `HU1 => FU1`
* Unknown extension `FU1 => HU1`

Does this match your thoughts and proposal?

akolotov (approved):

PR #514, opened 2021-02-23
Fix resend of stuck pending transactions

k1rill-fedoseev:
This PR is to address behavior when due to congestion in the chain, the privileged oracles' transactions with 0 gasprice could getting stuck. Later, when the oracle is trying to resend them it will get error message that the transaction with the same hash is already in the txpool, and it will not be added back to the resend queue for further check if it was eventually mined. 
As the result another transaction of the same oracle with the same nonce will re-write the original transaction and the bridge message will not be transferred at all.

### DIFF
--- a/oracle/src/sender.js
+++ b/oracle/src/sender.js
@@ -174,16 +174,17 @@ async function main({ msg, ackMsg, nackMsg, channel, scheduleForRetry, scheduleT
           e.message
         )
 
-        if (isResend) {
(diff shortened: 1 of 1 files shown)


thebeanespinoza83 (approved):

thebeanespinoza83:
oracle/src/sender.js

akolotov (approved):

PR #530, opened 2021-03-13
Fallback eth_getLogs to explorer API for BSC

### LINKED ISSUE #524: Confirmations collected but there is no "Execute" button
The following behavior was discovered on ALM between BSC and xDAI for a transfer from the xDai chain:

image

Although the confirmation has been collected, the "Execute" button did not appear.

Link to ALM: https://alm-bsc-xdai.herokuapp.com/100/0x2215dfd050dbe89db1823b8a31d0216dfcc9324206f2f2de7a5757b54cd8fb90

### LINKED ISSUE #529: ALM does not work with BSC public nodes
Some operation in AMB Live Monitoring tool assumes that the RPC node it requests the bridge info from is able to provide transactions logs from the block `1`. But obviously it is not needed for ALM operations in the worst case it makes sense to request information starting for the block where the corresponding contract were deployed at. But even with this improvement some nodes will still fail: for example, the BSC public nodes return the error "exceed maximum block range: 5000". 

Consider to optimize the work with historical data in ALM to make it more robust.

k1rill-fedoseev:
Closes #524 
Closes #529

### DIFF
--- a/alm/Dockerfile
+++ b/alm/Dockerfile
@@ -19,7 +19,7 @@ COPY --from=contracts /mono/contracts/build ./contracts/build
 COPY commons/package.json ./commons/
 COPY alm/package.json ./alm/
 COPY yarn.lock .
-RUN NOYARNPOSTINSTALL=1 yarn install --frozen-lockfile --production
+RUN NOYARNPOSTINSTALL=1 yarn install --frozen-lockfile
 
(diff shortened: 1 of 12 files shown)


akolotov (commented):
Tried to test your fix and got the following:

image

akolotov (approved):

PR #555, opened 2021-04-21
Fix/add rpc block limit

leonprou:
Initially indented to open this on our fork. But would be pleased if you get a look guys
Making `eth_getLogs` query over a big span of blocks is problematic for many RPC's on the BSC network. This introduces a limit to fetch the events by chunks.


k1rill-fedoseev:
Thanks, @leonprou, that is a really nice catch.
I have refactored your changes and applied them in a separate PR #557.
Please, take a look and we will merge it.

leonprou:
Thanks! Sure will take a look

PR #557, opened 2021-04-23
Add blocks range limit for logs polling

k1rill-fedoseev:
Addresses the issue found under #555 

Polling events from the unbounded block range can be a source of several problems:
* Some RPC endpoints have a strict limit for the maximum range limit for `eth_getLogs` (e.g. standard BSC JSON-RPCs)
* If requested block range contained lots of events, oracle won't be able to handle all of them in the fixed amount of time (defined by `ORACLE_MAX_PROCESSING_TIME`). It will fail to make any progress from such point.

This PR fixes both problems by introducing a configurable range size limit for both sides.


akolotov (approved):

akolotov:
@leonprou please take a look

leonprou:
@k1rill-fedoseev looks good!

PR #580, opened 2021-06-13
Add safe-execute button in ALM

### LINKED ISSUE #551: Use safeExecuteSignaturesWithAutoGasLimit in ALM
Since a more convenient way (for ordinary users) to execute AMB messages on the Foreign side has been [recently](https://github.com/poanetwork/tokenbridge-contracts/pull/587) implemented in the AMB contracts, it makes sense to start using a new `safeExecuteSignaturesWithAutoGasLimit` method instead of `executeSignatures` in AMB Live Monitoring app. It will allow users to re-send the claim transaction if by some reason the previous attempt failed due to the issue in the token contract (e.g. a token contract uses an AMM to perform some exchange but the slippage is not suitable). And also it provides more accurate gas estimate for the users. Here is examples:
  * Kovan-Sokol AMB:
    * usage of [`executeSignatures` leads](https://kovan.etherscan.io/tx/0x2c1def8f7a2cf8c80d485990172c4f6ff8681b35e13bf10529629ec7803bfab3) to the estimate in `1'133'157` gas.
    * usage of [`safeExecuteSignaturesWithAutoGasLimit` leads](https://kovan.etherscan.io/tx/0x320bab625c44b6d424dd6ba60006ad9b8e53747cf5b73a9dfba5a94c2d9c45b5) to the estimate in `334'060` gas.

  * BSC-xDAI AMB:
    * usage of [`executeSignatures` leads](https://bscscan.com/tx/0x195fcd856c4fc0a3a06d78b99d5de5a5eab272c3c72fca8a640886aa86d6225e) to the estimate in `1'572'403` gas.
    * usage of [`safeExecuteSignaturesWithAutoGasLimit` leads](https://bscscan.com/tx/0x3ae4a9135fd20b7d712a476c07c573876f79d61becf0ae3b13d26dd3bf890ae8) to the estimate in `353'467` gas.

In order to distinguish which method must be called, consider the return value of the method `getBridgeInterfacesVersion` of the AMB contract on the Foreign side: if it returns a value below `5.7.0`, the method `executeSignatures` must be used since it is an old version of AMB contract, if it is `5.7.0` or higher, the method `safeExecuteSignaturesWithAutoGasLimit` must be used.

### LINKED ISSUE #573: execute message by safeExecuteSignaturesWithAutoGasLimit in ALM
Consider to modify behavior of ALM as so if a user press the "Execute" button, the application checks whether the Foreign AMB contract supports the `safeExecuteSignaturesWithAutoGasLimit` method (e.g. by comparison of the version of the contract) and uses this method instead of `executeSignatures`.

k1rill-fedoseev:
Closes #573 
Closes #551 

Screenshot 2021-06-12 at 22 36 56

### DIFF
--- a/alm/src/components/ManualExecutionButton.tsx
+++ b/alm/src/components/ManualExecutionButton.tsx
@@ -25,6 +25,7 @@ const StyledButton = styled.button`
 `
 
 interface ManualExecutionButtonParams {
+  safe?: boolean
   messageData: string
   setExecutionData: Function
   signatureCollected: string[]
@@ -32,6 +33,7 @@ interface ManualExecutionButtonParams {
 }
 
 export const ManualExecutionButton = ({
+  safe,
   messageData,
   setExecutionData,
   signatureCollected,
@@ -72,7 +74,10 @@ export const ManualExecutionButton = ({
       const signatures = packSignatures(signatureCollected.map(signatureToVRS))
       const messageId = messageData.slice(0, 66)
       const bridge = foreign.bridgeContract
-      const data = bridge.methods.executeSignatures(messageData, signatures).encodeABI()
+      const executeMethod = safe
+        ? bridge.methods.safeExecuteSignaturesWithAutoGasLimit
+        : bridge.methods.executeSignatures
+      const data = executeMethod(messageData, signatures).encodeABI()
       setManualExecution(false)
 
       library.eth
@@ -132,14 +137,15 @@ export const ManualExecutionButton = ({
       messageData,
       signatureCollected,
       setExecutionData,
-      setPendingExecution
+      setPendingExecution,
+      safe
     ]
   )
 
   return (
     <div className="is-center">
       <StyledButton className="button outline" onClick={() => setManualExecution(true)}>
-        Execute
+        {safe ? 'Safe Execute' : 'Execute'}
       </StyledButton>
     </div>
   )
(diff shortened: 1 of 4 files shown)


akolotov (changes requested):
It will not obvious for end users which button to press. So, consider to have only one button which will always invoke safe-execution on the new version of contracts. For the previous version of the AMB contracts it should call the method `executeSignatures`.

akolotov:
> HI think there might be some rare situations, in which message should be recorded as failed.

May you suggest formal criteria how to decide whether the message must be recorded as failed? I would introduce a check box near to the button and output these criteria when a user moves the cursor above the check box. If the check box is chosen `executeSignatures` is used. It will make the user expectation more obvious.

k1rill-fedoseev:
Screenshot 2021-06-14 at 18 01 33

Do you think such UI would be appropriate?

alm/src/components/ManualExecutionButton.tsx
```diff
@@ -132,15 +141,35 @@ export const ManualExecutionButton = ({
...
+          className="is-center"
+          style={{ paddingTop: 10 }}
+        >
+          <input
+            type="checkbox"
+            id="allow-failures"
+            checked={allowFailures}
+            onChange={e => setAllowFailures(e.target.checked)}
+          />
+          <label htmlFor="allow-failures">Allow Failures</label>
```
akolotov, line 170: Does it make sense to call it "Unsafe execution" as so only users which understand what it means consider to use it?

k1rill-fedoseev: Ok, we might do so. I have renamed it to "Unsafe mode".

akolotov (approved):

PR #582, opened 2021-07-01
Improve confirm-relay feature

### LINKED ISSUE #569: Confirmation for multiple skipped transfer requests
It could be a situation when the oracle skipped messages to relay, for example, due to instability of the JSON RPC nodes. It could be fixed either by resetting the revisited block number or by using the [`confirmRelay`](https://github.com/poanetwork/tokenbridge/blob/master/oracle/src/confirmRelay.js) functionality which is more recommended since reduce load on the JSON RPC facilities.

Consider to improve `confirmRelay` to provide an ability to specify the list of skipped messages (by transactions hashes).

k1rill-fedoseev:
This PR incorporates the following set of changes to the confirm-relay script.

* Add missing log issuing contract address check in `getEventsFromTx`
* Add more npm/yarn scripts for processing different event types
* Allow to process multiple input transactions via confirm-relay script (e.g. `yarn confirm:transfer <txHash1> <txHash2>`). Closes #569 
* Add e2e test for simple AMB confirm-relay operations
* Fix minor logging issues


oracle/config/transfer-watcher.config.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
   },
   event: 'Transfer',
   eventFilter: { to: process.env.COMMON_FOREIGN_BRIDGE_ADDRESS },
+  sender: 'home',
```
akolotov, line 33: should it be the foreign side here? The oracle listens `Transfer` events of an erc20 token to send the confirmations to the home side

oracle/scripts/start-worker.sh
```diff
@@ -8,12 +8,12 @@ LOGS_DIR="logs/"
 WORKER="${WORKERS_DIR}${1}.js"
 CONFIG="${2}.config.js"
 LOG="${LOGS_DIR}${2}.txt"
-TX_HASH="${3}"
+TX_HASH=${@:3}
```
akolotov, line 11: Imagine that we have dozens of unexecuted messages, it makes sense to specify them by a file rather than by a list of parameters.

akolotov (changes requested):

oracle/config/transfer-watcher.config.js
k1rill-fedoseev: As you said, confirmation is send on the home side, that's why the sender should be configured to `home`. The only place where it is set to `foreign`, is for a collected-signatures watcher.

oracle/scripts/start-worker.sh
k1rill-fedoseev: Will it be easy to make such file on a running oracle inside docker container. I assumed that confirm-relay script is usually executed via `docker-compose exec`/`docker-compose run`, so passing transaction hashes via command line seemed the easiest way for me.

We can also do something like:
```
docker-compose exec ... confirm-transfer $(cat ./tx_hashes.txt)
```

oracle/config/transfer-watcher.config.js
akolotov: Right. Missed that it was for the sender.

oracle/scripts/start-worker.sh
akolotov: Is there any limitation to pass data through the command line to avoid receiving `Argument list too long` by using the approach you suggested? Specifying the transactions hashes by a file will not meet such issue at all.

k1rill-fedoseev: Ok, seems that it is possible to accept both, direct hashes and files path at the same time in the CLI. I have updated the code in that way.

akolotov (approved):

PR #583, opened 2021-07-02
Fix logging in gas price service

### LINKED ISSUE #552: Don't report that Gas Price API is not available if gas price feeder is not configured
Currently if the `COMMON_*_GAS_PRICE_SUPPLIER_URL` variable is not configured the oracle floods the logs with messages like this:

```
Apr 15 12:16:30 ... ...  {"level":50,"time":1618478190819,"msg":"Gas Price API is not available. Only absolute URLs are supported","name":"sender-foreign","module":"gasPrice","v":1}
Apr 15 12:16:30 ... ...  {"level":50,"time":1618478190919,"msg":"Gas Price API is not available. Only absolute URLs are supported","name":"sender-home","module":"gasPrice","v":1}
```

Obviously there is no need to report about the issue after exceeding every `ORACLE_*_GAS_PRICE_UPDATE_INTERVAL` timeout, it is enough to report that gas price feeder is not used only once when the corresponding sender starts -- similar like the message "You are using http (...) on undefined network. In production https must be used instead." is logged.

k1rill-fedoseev:
Closes #552


oracle/src/tx/test2.js
```diff
@@ -0,0 +1,20 @@
+const Web3 = require('web3')
```
akolotov, line 1: Is it required to have a module required for testing in `oracle/src/tx`?

k1rill-fedoseev: I have committed this file by mistake. It is not needed.

PR #587, opened 2021-07-11
Async calls error codes

k1rill-fedoseev:
This PR reworks the way of error handling in the async calls watcher.
It also adds a strict limit on the returned data length, fixes eth_call revert handling and limits upper gas limit bound for async eth_call.


oracle/src/events/processAMBInformationRequests/calls/ethGetStorageAt.js
```diff
@@ -12,7 +14,7 @@ async function callArchive(web3, data, foreignBlock) {
   const { 0: address, 1: slot, 2: blockNumber } = web3.eth.abi.decodeParameters(['address', 'bytes32', 'uint256'], data)
 
   if (toBN(blockNumber).gt(toBN(foreignBlock.number))) {
-    return [false, '0x']
+    return [false, ASYNC_CALL_ERRORS.BLOCK_IS_IN_THE_FUTURE]
```
akolotov, line 17: Does it also make sense to handle the situation when there is no storage for an account at all (for EOA accounts)?

oracle/src/utils/constants.js
```diff
@@ -27,5 +27,17 @@ module.exports = {
...
+    NOT_FOUND: '0x0000000000000000000000000000000000000000000000000000000000000000',
+    // requested custom block does not exist yet or its timestamp is greater than the home block timestamp
+    BLOCK_IS_IN_THE_FUTURE: '0x0000000000000000000000000000000000000000000000000000000000000001',
+    // eth_call has reverted or finished with OOG error
+    REVERT: '0x0000000000000000000000000000000000000000000000000000000000000002',
+    // evaluated output length exceeds allowed length of 64 KB
+    RESULT_IS_TOO_LONG: '0x0000000000000000000000000000000000000000000000000000000000000003'
+  },
+  MAX_ASYNC_CALL_RESULT_LENGTH: 64 * 1024,
+  ASYNC_ETH_CALL_MAX_GAS_LIMIT: 100000000
```
akolotov, line 42: Do you think to have a lower gas limit will allow to avoid ddos attacks on the RPC nodes? Since several oracles will requests the same request at the same time it could insult the nodes.

k1rill-fedoseev: I know that Infura uses 10x latest block gas limit as a cap for eth_call/eth_estimateGas methods in order to prevent such kind of abuse. So, I think we are quite safe to go with 100.000.000 gas limit.

oracle/src/events/processAMBInformationRequests/calls/ethGetStorageAt.js
k1rill-fedoseev: Ethereum clients usually does not distinguish such cases at all. There is no hardcoded restrictions there, which does not allow for EOA to have non-zero values in its storage. They are simply all-zero values, since there is no way to initialise them.

Also, JSON RPC returns all-zeros value for any eth_getStorageAt call related to EOA, so we are just copying the same behaviour.

PR #588, opened 2021-07-12
Refactor/async call serializers

k1rill-fedoseev:
Add missing fields to the serialised block/transaction/receipt objects returned by async information calls.

### DIFF
--- a/oracle/src/events/processAMBInformationRequests/estimateGas.js
+++ b/oracle/src/events/processAMBInformationRequests/estimateGas.js
@@ -51,6 +51,20 @@ async function estimateGas({ web3, homeBridge, validatorContract, messageId, sta
       throw new InvalidValidatorError(`${address} is not a validator`)
     }
 
+    logger.debug('Check if InformationRetrieved event for this message already exists')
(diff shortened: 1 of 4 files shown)


oracle/src/events/processAMBInformationRequests/estimateGas.js
```diff
@@ -51,6 +51,20 @@ async function estimateGas({ web3, homeBridge, validatorContract, messageId, sta
       throw new InvalidValidatorError(`${address} is not a validator`)
     }
 
+    logger.debug('Check if InformationRetrieved event for this message already exists')
+    const logs = await homeBridge.getPastEvents('InformationRetrieved', {
+      fromBlock: '0',
```
akolotov, line 56: does it make sense to use the deployed block number stored in the contract as the start block for the filter or even the block where the corresponding `UserRequestForInformation` was raised?

k1rill-fedoseev: I like the idea to use blockNumber from UserRequestForInformation event. Applied it in the latest commit.

PR #613, opened 2021-10-07
Include cDAI balance in balanceDiff

k1rill-fedoseev:


monitor/getBalances.js
```diff
@@ -66,7 +66,13 @@ async function main(bridgeMode, eventsInfo) {
     const mintedCoinsBN = new BN(mintedCoins)
     const burntCoinsBN = new BN(burntCoins)
     const totalSupplyBN = mintedCoinsBN.minus(burntCoinsBN)
-    const foreignErc20BalanceBN = new BN(foreignErc20Balance).plus(lateForeignConfirmationsTotalValue)
+    let foreignErc20BalanceBN = new BN(foreignErc20Balance).plus(lateForeignConfirmationsTotalValue)
+    try {
+      const invested = await foreignBridge.methods.investedAmount(erc20Address).call()
```
akolotov, line 71: Does it make sense to set `foreignDelayedBlockNumber` to get the value at a specific moment of time?

k1rill-fedoseev: Yes, it makes total sense. Thanks.

PR #631, opened 2021-12-17
Add support for EIP1559 gas price oracle

k1rill-fedoseev:


oracle/src/confirmRelay.js
```diff
@@ -153,11 +153,11 @@ async function main({ sendJob, txHashes }) {
 }
 
 async function sendJobTx(jobs) {
-  await GasPrice.start(chain, true)
-  const gasPrice = GasPrice.getPrice().toString(10)
-
   const { web3 } = config.sender === 'foreign' ? config.foreign : config.home
 
+  await GasPrice.start(chain, web3, true)
```
akolotov, line 158: Will `mycrypto/gas-estimation` work if specify several URLs in the RPC URL parameter?

commons/utils.js
```diff
@@ -176,12 +177,20 @@ const gasPriceWithinLimits = (gasPrice, limits) => {
...
   gasPrice = gasPriceWithinLimits(gasPrice, limits)
-  return toBN(toWei(gasPrice.toFixed(2).toString(), 'gwei'))
+  return toWei(gasPrice.toFixed(2).toString(), 'gwei')
 }
 
-const gasPriceFromSupplier = async (url, options = {}) => {
+const gasPriceFromSupplier = async (web3, url, options = {}) => {
   try {
     let json
+    if (url === 'eip1559-gas-estimation') {
```
akolotov, line 186: Does it make sense to reflect this way to configure the gas price supplier [here](https://github.com/poanetwork/tokenbridge/blob/master/CONFIGURATION.md)?

oracle/src/confirmRelay.js
k1rill-fedoseev: It should, all RPC calls are going through the same created web3 object, so any url fallback logic should work here as well.

PR #634, opened 2022-02-01
CollectedSignatures AMB watcher for MEV bundling

k1rill-fedoseev:


oracle/src/mevSender.js
```diff
@@ -0,0 +1,159 @@
...
+  setTimeout(runMain, config.pollingInterval)
+}
+
+async function main() {
+  try {
+    const jobs = Object.values(await redis.hgetall(config.mevJobsRedisKey)).map(JSON.parse)
+    const totalJobs = jobs.length
+
+    if (totalJobs === 0) {
+      logger.info('Nothing to process')
```
akolotov, line 64: Seems that this message will flood the logs. Does it make sense to make use the "debug" level for it?

oracle/src/events/processAMBCollectedSignaturesMEV/index.js
```diff
@@ -0,0 +1,174 @@
...
+              return
+            }
+            if (blockList.includes(sender)) {
+              logger.info({ sender }, 'Validator skips a message. Sender address is in the block list.')
+              return
+            }
+          }
+        }
+
+        logger.debug({ NumberOfCollectedSignatures }, 'Number of signatures to get')
```
akolotov, line 68: It make sense to check if the message was marked to go through the manual lane like it was done here: https://github.com/poanetwork/tokenbridge/blob/b17fff2b5678adf0272329870d7ade8927f16696/oracle/src/events/processAMBCollectedSignatures/index.js#L79-L85

oracle/src/mevSender.js
```diff
@@ -0,0 +1,159 @@
...
+
+    const maxProfit = await estimateProfit(
+      mevForeign.contract,
+      mevForeign.minGasPrice,
+      bestJob.executeData,
+      bestJob.value
+    )
+
+    if (maxProfit === '0') {
+      jobLogger.info(`No MEV opportunity found when testing with min gas price ${mevForeign.minGasPrice}, removing job`)
```
akolotov, line 86: In my tests I've got the following behavior:
```
Feb  6 16:13:29 amb-bsc oracle_bridge_alpha_mevSender_1/6a29bca10ef1[2320]: {"level":30,"time":1644164009281,"msg":"Sending MEV bundles","validator":"0xACeFD9907F5D4d7Fc387193414733a111863c8D6","name":"mev-sender-foreign","eventTransactionHash":"0x2b6c8df5db39f90c0d23ad0512e7482bc15a445f3315bbddccb12993c0eb9554","nonce":4431,"fromBlock":15023194,"toBlock":15023198,"v":1}
Feb  6 16:13:29 amb-bsc oracle_bridge_alpha_mevSender_1/6a29bca10ef1[2320]: {"level":30,"time":1644164009775,"msg":"Tx generated 0xfdeec6ee313005977991457fe00b24e0b8ed8f8f0dbe89487b1f5595c4ddb1fc for event Tx 0x2b6c8df5db39f90c0d23ad0512e7482bc15a445f3315bbddccb12993c0eb9554","validator":"0xACeFD9907F5D4d7Fc387193414733a111863c8D6","name":"mev-sender-foreign","eventTransactionHash":"0x2b6c8df5db39f90c0d23ad0512e7482bc15a445f3315bbddccb12993c0eb9554","txHash":"0xfdeec6ee313005977991457fe00b24e0b8ed8f8f0dbe89487b1f5595c4ddb1fc","v":1}
Feb  6 16:13:45 amb-bsc oracle_bridge_alpha_mevSender_1/6a29bca10ef1[2320]: {"level":40,"time":1644164025014,"msg":"Gas price API is not configured, will fallback to the contract-supplied gas price","validator":"0xACeFD9907F5D4d7Fc387193414733a111863c8D6","name":"mev-sender-foreign","module":"gasPrice","chainId":"foreign","v":1}
Feb  6 16:13:45 amb-bsc oracle_bridge_alpha_mevSender_1/6a29bca10ef1[2320]: {"level":30,"time":1644164025714,"msg":"No MEV opportunity found when testing with min gas price 5000000000, removing job","validator":"0xACeFD9907F5D4d7Fc387193414733a111863c8D6","name":"mev-sender-foreign","eventTransactionHash":"0x2b6c8df5db39f90c0d23ad0512e7482bc15a445f3315bbddccb12993c0eb9554","v":1}
```
As you can see it is only 16 seconds between two actions: the transaction was sent, the corresponding job was removed. Isn't it too fast? May we have larger delay for checks whether transactions was executed successfully? On the Ethereum Mainnet it could take minutes before the transaction is included into the chain so the sender could send several transactions for the same job in short period of time.

akolotov (changes requested):

oracle/src/events/processAMBCollectedSignaturesMEV/index.js
k1rill-fedoseev: I think that it is useful to consider manual messages, since it makes sense to the oracle to execute them as well, if they are profitable for it.

oracle/src/mevSender.js
k1rill-fedoseev: Well, the job is not removed immediately after it was sent to the Flashbots RPC. Indeed, there is no guarantee that it will be included at all, since we send request only for a few upcoming blocks.

After the job tx was sent to the RPC, the oracle will then again reinspect it on the next iteration, and will remove it only if it is no longer profitable (e.g. if it was already executed by oracle/someone else). If the transaction is still profitable, it will continue to send it, until it become non-profitable (e.g. when message was successfully executed).

oracle/src/mevSender.js
k1rill-fedoseev: Ok, no problem

oracle/src/mevSender.js
akolotov: Ok. Understood. So does it make sense to make the polling interval independent from ORACLE_FOREIGN_RPC_POLLING_INTERVAL? So, we can control how fast the sender re-estimates the transaction?

oracle/src/events/processAMBCollectedSignaturesMEV/index.js
akolotov: I think we need to be consistent here:
- If the allowance or block lists are configured we skips the message even if it could be profitable.
- But if the message is sent to the manual line by the contract we try to estimate if it is profitable or not. 

What if we check whether the sender/receiver of the message is within the allowance list and only if it is true, check the lane the message was sent through? Or at least to make the second check configurable. I just would like to have an ability to not estimate profit for messages I do not intend to execute.

oracle/src/mevSender.js
k1rill-fedoseev: Well, since in the mevSender, ORACLE_FOREIGN_RPC_POLLING_INTERVAL is not used for anything else, there is no need to introduce another variable.

For example, we can `ORACLE_FOREIGN_RPC_POLLING_INTERVAL=15000` in the watcher and `ORACLE_FOREIGN_RPC_POLLING_INTERVAL=70000` in the sender.

Also, note that since sender will send MEV bundles for 5 upcoming blocks by default, there is no reason to set polling interval lower than `5*avg_block_time ~= 70 seconds`.

akolotov: Got you. In this case consider to have `ORACLE_FOREIGN_RPC_POLLING_INTERVAL` configured for mev-sender in `oracle/docker-compose-helpers.yml` with the corresponding comment. I would also say that we need to re-define `ORACLE_HOME_START_BLOCK` in the docker-compose file as a start point for watching the messages to execute.

oracle/docker-compose-helpers.yml
```diff
@@ -13,3 +21,35 @@ services:
...
+    restart: unless-stopped
+    entrypoint: yarn mev:watcher:collected-signatures
+  mevSender:
+    cpus: 0.1
+    mem_limit: 500m
+    image: poanetwork/tokenbridge-oracle:latest
+    env_file: ./.env
+    environment:
+      NODE_ENV: production
+      ORACLE_VALIDATOR_ADDRESS: ${ORACLE_VALIDATOR_ADDRESS}
```
akolotov, line 46: Should it contain also
```
ORACLE_VALIDATOR_ADDRESS_PRIVATE_KEY: ${ORACLE_VALIDATOR_ADDRESS_PRIVATE_KEY}
```
?

akolotov (approved):

PR #641, opened 2022-03-05
ALM: Do not show failed confirmations for successfully processed messages

k1rill-fedoseev:


akolotov (commented):
Here is the requirement for this ALM enhancement: https://github.com/poanetwork/tokenbridge/issues/642. Please pay attention that Success is changed to Confirmed and how meaning of Fail was changed.

akolotov (approved):

PR #651, opened 2022-04-07
Missing case when check nonce error

cuonghx-gutech:
I found a missing case when checking the nonce error.

Error: "The tx doesn't have the correct nonce" not contained in isNonceError function so the nonce in Redis does not update so it is a loop error

ca26335d9543c0983b864d32f47d0026

### DIFF
--- a/oracle/src/utils/utils.js
+++ b/oracle/src/utils/utils.js
@@ -169,7 +169,8 @@ function isNonceError(e) {
     message.includes('transaction nonce is too low') ||
     message.includes('nonce too low') ||
     message.includes('transaction with same nonce in the queue') ||
-    message.includes('oldnonce')
+    message.includes('oldnonce') ||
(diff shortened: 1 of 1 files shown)


k1rill-fedoseev (approved):
Just to confirm, have you used Ganache in your environment? I didn't find this error in any other clients.

cuonghx-gutech:
I using my own company network and ropsten

cuonghx-gutech:
Yeah. May because Ganache, I using ganache at first after that I switch to geth network(our network). May I ask you why happened at Ganache? And then you can close this pull

k1rill-fedoseev:
Different clients use different error messages, and while there are no uniformly accepted error codes for them, we need to stick to some particular error strings in our logic.
Here we have missed the error string that is exclusively used in Ganache, so thank you for adding that one.

